PR #22, opened 2023-01-04
Add content_length to WSGI environ

shulcsm:
According to PEP333 `CONTENT_LENGTH` variable is optional but it also says:
> The application should not attempt to read more data than is specified by the CONTENT_LENGTH variable.  

So most implementations assume length is 0 if header is missing.
See for details: https://wsgi.readthedocs.io/en/latest/proposals-2.0.html#unknown-length-wsgi-input

As for the fix itself i'm not sure if is correct. Could just grab length in python?

### DIFF
--- a/granian/_granian.pyi
+++ b/granian/_granian.pyi
@@ -78,3 +78,4 @@ class WSGIScope:
     query_string: str
     headers: Dict[str, str]
     body: bytes
+    length: int
--- a/granian/wsgi.py
+++ b/granian/wsgi.py
@@ -44,7 +44,8 @@ def _callback_wrapper(callback):
             'QUERY_STRING': scope.query_string,
             'REMOTE_ADDR': scope.client,
             'wsgi.url_scheme': scope.scheme,
-            'wsgi.input': scope.body
+            'wsgi.input': scope.body,
+            "CONTENT_LENGTH": scope.length
         }
         resp = Response()
 
--- a/src/wsgi/types.rs
+++ b/src/wsgi/types.rs
@@ -1,10 +1,9 @@
-use hyper::{Body, Uri, Request, body::Bytes};
+use hyper::{body::Bytes, Body, Request, Uri};
 use pyo3::prelude::*;
 use pyo3::types::PyBytes;
 use std::{collections::HashMap, net::SocketAddr};
 
-
-#[pyclass(module="granian._granian")]
+#[pyclass(module = "granian._granian")]
 pub(crate) struct WSGIScope {
     #[pyo3(get)]
     scheme: String,
@@ -17,7 +16,9 @@ pub(crate) struct WSGIScope {
     client: String,
     #[pyo3(get)]
     headers: HashMap<String, String>,
-    body: Bytes
+    body: Bytes,
+    #[pyo3(get)]
+    length: usize,
 }
 
 impl WSGIScope {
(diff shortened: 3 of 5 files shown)


cirospaciari:
@gi0baro actually the problem is HTTP_CONTENT_LENGTH should be CONTENT_LENGTH and HTTP_CONTENT_TYPE should be HTTP_CONTENT_TYPE, i have this exactly same issue

cirospaciari:
@gi0baro another thing wsgi.input cannot be bytes, in my case I used BytesIO but is also wrong you can test with:

```python
from flask import Flask, make_response, request
from socketify import WSGI

app = Flask(__name__)

@app.route('/')
def index():
    """Test 6: Plaintext"""
    response = make_response(b"Hello, World!")
    response.content_type = "text/plain"
    return response

@app.route('/form', methods=['POST'])
def post_test():
    """Test 6: Plaintext"""
    user = request.form["name"]
    response = make_response(f"Hello, World {user}!")
    response.content_type = "text/plain"
    return response
@app.route('/json', methods=['POST'])
def post_test():
    """Test 6: Plaintext"""
    data = request.json
    response = make_response(f"Hello, World {data}!")
    response.content_type = "text/plain"
    return response

@app.route('/raw', methods=['POST'])
def post_test():
    """Test 6: Plaintext"""
    data = request.data
    response = make_response(f"Hello, World {data}!")
    response.content_type = "text/plain"
    return response
```

I fixed socketify by taking a look at https://github.com/benoitc/gunicorn/blob/2f9eb19b666868d324150072e7489788688c91cb/gunicorn/http/body.py

and I come up with https://github.com/cirospaciari/socketify.py/blob/392af9784d6a94403e58d2c02652aedc3379bf4c/src/socketify/wsgi.py#L25

shulcsm:
> @gi0baro actually the problem is HTTP_CONTENT_LENGTH should be CONTENT_LENGTH and HTTP_CONTENT_TYPE should be CONTENT_TYPE, i have this exactly same issue

I'm not sure about incoming hyper request semantics but it probably supports chunked-encoding?  I'm also not sure if you can just trust and proxy the header.

cirospaciari:
> > @gi0baro actually the problem is HTTP_CONTENT_LENGTH should be CONTENT_LENGTH and HTTP_CONTENT_TYPE should be CONTENT_TYPE, i have this exactly same issue
> 
> I'm not sure about incoming hyper request semantics but it probably supports chunked-encoding? I'm also not sure if you can just trust and proxy the header.

I don't know where you are getting content_length from (in the case of the header) but you can override to always be the real body length (must be a string)

shulcsm:
Right so at least the string part seems wrong. But `HTTP_CONTENT_LENGTH` is raw header from request (might be missing) and `CONTENT_LENGTH` is the actual length of buffered body?

cirospaciari:
> Right so at least the string part seems wrong. But `HTTP_CONTENT_LENGTH` is raw header from request (might be missing) and `CONTENT_LENGTH` is the actual length of buffered body?

checking guvicorn, only CONTENT_LENGTH and CONTENT_TYPE is provided, no HTTP_CONTENT_LENGTH or HTTP_CONTENT_TYPE counter part.

Flask also expects CONTENT_LENGTH and CONTENT_TYPE without HTTP_ prefix

shulcsm:
`HTTP_` headers are not relevant to implementation of WSGI app/request but should be forwared in raw form.
> Variables corresponding to the client-supplied HTTP request headers (i.e., variables whose names begin with "HTTP_"). The presence or absence of these variables should correspond with the presence or absence of the appropriate HTTP header in the request.

cirospaciari:
> `HTTP_` headers are not relevant to implementation of WSGI app/request but should be forwared in raw form.
> > Variables corresponding to the client-supplied HTTP request headers (i.e., variables whose names begin with "HTTP_"). The presence or absence of these variables should correspond with the presence or absence of the appropriate HTTP header in the request.
> 

Just saying what gunicorn is doing and what flask is expecting to work with, I followed the specs for socketify.py WSGI and come up with simular solution (but I used BytesIO instead of bytes because the spec says to use stream like objects), and to post work properly we need to provide CONTENT_LENGTH and CONTENT_TYPE, not only HTTP_ prefixed counter parts (you can test with flask, falcon, Django etc)

cirospaciari:
Also, https://wsgi.readthedocs.io/en/latest/proposals-2.0.html#unknown-length-wsgi-input are just proposals not the spec itself, but POST should work at least on Flask and Django (the most popular WSGI WebFrameworks) here is the actual spec for POST https://wsgi.readthedocs.io/en/latest/specifications/handling_post_forms.html

gi0baro:
@shulcsm @cirospaciari thank you for reporting this, I gonna commit directly in this PR to use the `gunicorn` approach regarding `CONTENT_TYPE` and `CONTENT_LENGTH`

gi0baro:
@shulcsm since I can't directly commit on your fork, I opened a PR on your fork for this, see https://github.com/shulcsm/granian/pull/1

shulcsm:
> @shulcsm since I can't directly commit on your fork, I opened a PR on your fork for this, see [shulcsm#1](https://github.com/shulcsm/granian/pull/1)
I'm not sure it's worth jumping the hoops. You can just close this PR.

Are you sure just garbing length from header will work? 
What happens with chunk-encoded requests? Even if hyper::Request always buffered i doubt it creates the header.

gi0baro:
> Are you sure just garbing length from header will work? What happens with chunk-encoded requests? Even if hyper::Request always buffered i doubt it creates the header.

This is what `gunicorn` does, so let's say I feel pretty confident is correct.
Eventually we can add a dedicated test for it.

PR #31, opened 2023-01-11
Performance improvement in futures handling

cirospaciari:
As said in the following discussions:
https://github.com/emmett-framework/granian/discussions/30

Almost sure this optimization will make granian ASGI faster than uvicorn+uvloop+httptools+gunicorn (in local tests is)

PS the same improvement can be made in add_done_callback in `src/runtime.rs`, `src/generic.rs` and `src/lib.rs`

### DIFF
--- a/granian/_futures.py
+++ b/granian/_futures.py
@@ -1,17 +1,12 @@
 import asyncio
 
+async def task_wrapper(task, watcher):
+    try:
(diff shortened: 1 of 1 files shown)


gi0baro:
Looks good! Thank you @cirospaciari

PR #63, opened 2023-02-16
add custom tasks

cirospaciari:
Testing ASGI:
Before custom Task: 93k req/s 
After custom Task: 113k req/s 

About 21% improvement, I did not test but RSGI should get a performance boost too.

Main differences in this Task:

- This class does not support current_task
- This class executes the first step like node.js Promise
- This class does not have inheritance (is a Future Like by itself)
- This class allows reusing the same instance again (check TaskFactory)

The use of TaskFactory in socketify for PyPy improves the performance a lot! (30% to 50%)
By executing the first step sync instead of async this Task reduces the overhead of async functions that executes sync tasks before the first await (basically the most used case) and for cases when the async function only actually does sync things (a lot of cases)

### DIFF
--- a/granian/_loops.py
+++ b/granian/_loops.py
@@ -2,6 +2,7 @@ import asyncio
 import os
 import signal
 import sys
+from ._tasks import create_task
 
 from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple
 
@@ -72,6 +73,8 @@ loops = BuilderRegistry()
 @loops.register('asyncio')
 def build_asyncio_loop():
     loop = asyncio.new_event_loop() if os.name != 'nt' else asyncio.ProactorEventLoop()
+    # this installs the custom task as default everywhere
+    loop.set_task_factory(lambda loop, coro, context=None: create_task(loop, coro, context=context))
     asyncio.set_event_loop(loop)
     return loop
 
@@ -80,6 +83,8 @@ def build_asyncio_loop():
 def build_uv_loop(uvloop):
     asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
     loop = asyncio.new_event_loop()
+    # this installs the custom task as default everywhere (yes is compatible with uvloop on CPython)
+    loop.set_task_factory(lambda loop, coro, context=None: create_task(loop, coro, context=context))
     asyncio.set_event_loop(loop)
     return loop
 
--- a/granian/_tasks.py
+++ b/granian/_tasks.py
@@ -1,0 +1,615 @@
+
+# Copyright 2023 Ciro Spaciari
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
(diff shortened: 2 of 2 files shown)


gi0baro:
@cirospaciari this is quite a lot of code! I will need some time to dig into it.

If I understood this correctly, the main difference between this and `asyncio.Task` is that you save a loop schedule invoking the `_step` directly, and also avoid to store task data for `current_task`. If that's the case, I would prefer a subclass of `Task` with just the overrides of the involved parts.

A part from that, I can't merge this directly, as it is broken on Python 3.7; and I'm also questioning myself where overriding the task creation in the loop is a good idea for something so low-level like Granian. I mean using `loop.set_task_factory` can mean only two things:
- preventing developers to do this in their frameworks/apps
- loose all the performance increment in case they do

I'm really fascinated by the performance improvement though, so I think a good alternative might be to wrap just the application callable in a custom task object, eventually written in Rust. But this is just an idea, I gonna keep this opened until I figure out the best option here.

Thank you for your contributions!

cirospaciari:
> 

Avoiding inheritance also boosts performance (being a plain future-like object).
Avoiding `current_task` is not a performance strategy; it will not work if a task calls another task inside, because `_step` may try to execute sync inside another `_step`.

In socketify, I used a custom create_task in requests directly and also used set_task_factory, so I guarantee the behavior for executing the first step in the request, but also allow custom set_task_factory from users (if some performance is lost, this is the developer's choice)

We can try to fix it to support 3.7 but, soon 3.7 will be dropped by many people. (is almost at the end of security updates)

https://devguide.python.org/versions/

Don't worry about merging, I just want to help to give some ideas, you can always check this code in the socketify repo if you need it. I am also always available on Twitter/discord if you need me.

gi0baro:
@cirospaciari a quick update on this:

I experimented quite a lot in the last weeks on tasks and Python loop enhancements, and while I still see room for improvements, I get quite unstable values in benchmarks. Since I want to release 0.3 shortly, I gonna keep this opened as a reference, and get back to the topic for 0.4.

cirospaciari:
@gi0baro no worries take our time, this will only remove overhead, so scaling the benchmark may help to see the difference, the benchmark will only show the difference if we have some sync work to do before the IO bound work. This is the same strategy used by JS Promises, there is no reason to delay it with call_soon and add more pressure to the event loop, if you have a lot of concurrencies will also show better the difference (because will be more pressure/tasks on the event loop). You may want to try on different kernel versions like:

[Debian bullseye (stable)](https://packages.debian.org/bullseye/linux-headers-amd64): 5.10.158-2
[Debian bookworm (next stable/testing)](https://packages.debian.org/bullseye/linux-headers-amd64): 6.1.12-1
TFB also started running on new Citrine (Ubuntu 22.04.2 / 5.15.0-67-generic) on March 3, 2023

The great thing here is PyPy Factory and the ability to reuse tasks that you have control (aka requests), this will show much more difference if used on PyPy.

I think we can do much better here in Tasks, for sure more analysis and study are needed, and we have a lot of room to improve.

gi0baro:
This is partially replaced by #76

PR #95, opened 2023-06-28
add package entrypoint

gabrielmbmb:
I've added the `__main__.py` file so `python -m granian` can be used.

### DIFF
--- a/granian/__main__.py
+++ b/granian/__main__.py
@@ -1,4 +1,3 @@
 from granian.cli import cli
 
-if __name__ == "__main__":
-    cli()
+cli()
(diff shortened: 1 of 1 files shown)


granian/__main__.py
```diff
@@ -0,0 +1,4 @@
+from granian.cli import cli
+
+if __name__ == "__main__":
+    cli()
```
gi0baro, line 4: Isn't the if redundant here? Probably this can be changed into just:

```suggestion
cli()
```

gi0baro (commented):

See my review

Issue #97, opened 2023-07-04
Support listening on UNIX domain sockets


fladi:
It would be nice if granian could also listen on UNIX domain sockets (UDS), leveraging filesystem permissions to controll access. Both nginx and apache2 (and possible more webservers) support proxying to UDS.

ArtemIsmagilov:
To add a listen unix socket or file descriptor as in gunicorn for example https://docs.gunicorn.org/en/stable/settings.html#bind it is necessary to remove the `--host` , `--port` parameters and add `--bind` param.

This will be a critical change. Are you considering unix socket and file descriptor support as a listen server?

gi0baro:
> it is necessary to remove the --host , --port parameters and add --bind param.
> This will be a critical change.

@ArtemIsmagilov not necessarily. We can adapt `--host` to also parse `unix://` uri, or add an additional parameter `--unix` that will take precedence over `--host` if supplied.

> You will need to re-implement the handler depending on the passed parameter --bind . In Rust, we have an excellent strict parser that will determine which host, port, or path we want to use as a socket

That's not the point to enable support though; the whole Granian code nowadays expects a `TCPListener`, so the actual change to support this would be wrap that in an enum, add the Unix support and patch the code to match the actual listener type.

> Are you considering unix socket and file descriptor support as a listen server?
> Do you plan to use this approach? How do you see the addition of this feature?

To be honest, I see this as a _very low priority_ atm.
The only benefit for listening on unix sockets is by having Granian behind a reverse proxy on the same host, which is definitely a _niche use case_ and not necessarily something we want to invest time onto – also given enabling keepalive on the proxy would provide similar performance, and supporting such feature would need to have 2 different code paths due to Windows support.

ionelmc:
UDS are quite popular with uwsgi users: https://github.com/search?q=socket%3D%2F+uwsgi&type=code

I believe most [uwsgi users looking for an alternative](https://github.com/unbit/uwsgi/commit/5838086dd4490b8a55ff58fc0bf0f108caa4e079) would look for something with comparable performance or comparable feature parity, or both. You have the performance but not the feature parity.

This feature is as niche as gluing up a bunch of docker containers on the same host is - so is it really? One could even argue that granian is a niche performance server with just one way to run it, given how enormous uwsgi's feature set is...

At least give it some thought and outline how this could be implemented, or how much funding would change the priority on this.

gi0baro:
[...]

PR #118, opened 2023-09-03
The socket holder class seems to be unused, remove it

matthiask:
@gi0baro I have dabbled around in Rust during advent of code a bit, and am looking into using granian for hosting an ASGI app. I'm mostly using Unix sockets so I have been looking into #97 for a bit and reading the code. It seems to me that the Rust `SocketHolder` class is unused since its introduction but I probably missed something.


gi0baro:
@matthiask I can't remember why I kept it around.
So I gonna merge this, thank you!

PR #119, opened 2023-09-03
Add a development section to the README

matthiask:

PR #124, opened 2023-09-07
Align codestyle

### LINKED ISSUE #123: Python code formatting and linting
Considering the more people contributing to the project, especially having #116, will it be appropriate to establish the formatting and linting rules for Python code? The obvious benefit is that no contributor will doubt what development tools and styles to apply.

I will gladly provide a configuration for tools like `black`, `isort`, and `ruff`, for a start at least.

While doing #122, I deactivated my `black` and `ruff` plugins because any file save immediately changes too many things I didn’t touch. And those two are the industry standard by now.

gi0baro:
Close #123


matthiask (commented):
I think this would make a rework of #119 necessary, but that's definitely fine since I still have problems with maturin.

My personal preference is to not install formatters using `[project.optional-dependencies]` but only referencing them through the pre-commit configuration, for example https://github.com/matthiask/django-content-editor/blob/944b9b8b403ecdcfe520cd088569f52259f57fdd/.pre-commit-config.yaml#L25-L32 because I can reuse the installations between many projects and I do not have to install black / ruff / etc. once for each project, but whatever works. 

(I won't get into the specifics of code formatting, that's not an useful discussion. Only this: I try configuring tools as little as possible, but no less. The important part is actually using formatters and linters.)

pyproject.toml
Aeron, line 1: Ah, I see now. It came from the fellow [Pydantic’s configuration][pydantic]. I wasn’t sure why these particular choices.

I don’t know why Samuel and the contributors went against Black a little. But I don’t think we have any legacy constraints here to break Black’s rules.

As a different example, we could look at the [Structlog’s configuration][structlog]. It’s made differently in both Ruff and Black rules.

[pydantic]: https://github.com/pydantic/pydantic-core/blob/f6b14cce22b1539727ed0e0a47d2c680c79c1692/pyproject.toml#L100
[structlog]: https://github.com/hynek/structlog/blob/bb10cbbd007c222a55c7c1e63042fc2b3e683e1d/pyproject.toml#L111

We can do something in the middle.

Also, we can expand optional dependencies to give a choice to install linters or do it another way.

pyproject.toml
```diff
@@ -39,9 +39,11 @@ dependencies = [
 
 [project.optional-dependencies]
 test = [
+    "black~=23.7.0",
     "httpx~=0.23.0",
     "pytest~=7.1.2",
     "pytest-asyncio~=0.18.3",
+    "ruff~=0.0.287",
     "websockets~=10.3"
 ]
```
Aeron, line 48: I would propose to have it like the following:

```suggestion
[project.optional-dependencies]
test = [
    "httpx~=0.23.0",
    "pytest~=7.1.2",
    "pytest-asyncio~=0.18.3",
    "websockets~=10.3"
]
lint = [
    "black~=23.7.0",
    "ruff~=0.0.287",
    "mypy~=1.5.1"
]
dev = ["granian[test,lint]"]
```

Having the `dev` is optional, of course. It’s just a convenience.

pyproject.toml
```diff
@@ -61,5 +63,20 @@ build-backend = "maturin"
 module-name = "granian._granian"
 bindings = "pyo3"
 
+[tool.ruff]
+line-length = 120
+extend-select = ['Q', 'RUF100', 'C90', 'I']
+extend-ignore = ['E501']
+flake8-quotes = { inline-quotes = 'single', multiline-quotes = 'double' }
+mccabe = { max-complexity = 13 }
+isort = { combine-as-imports = true, lines-after-imports = 2, known-first-party = ['granian', 'tests'] }
```
Aeron, line 72: I would suggest doing it more meticulously than just selecting everything and ignoring a few.

Something like the following:

```suggestion
[tool.ruff]
line-length = 120
extend-select = ['Q', 'RUF100', 'C90', 'I']
extend-select = [
    # E and F are enabled by default
    "B",     # flake8-bugbear
    "C4",    # flake8-comprehensions
    "C90",   # mccabe
    "D",     # pydocstyle
    "RUF100" # ruff (unused noqa)
    "I",     # isort
    "N",     # pep8-naming
    "S",     # flake8-bandit
    "W",     # pycodestyle
]
ignore = [
    "D1",   # public-* (use your best judgement)
    "D203", # one-blank-line-before-class
    "D213", # multi-line-summary-second-line
    "D215", # section-underline-not-over-indented
    "D400", # ends-in-period (in favour of D415)
    "D401", # non-imperative-mood
    "D404", # no-this-prefix
    "D406", # new-line-after-section-name
    "D407", # dashed-underline-after-section
    "E501", # line-too-long (black decides)
]

[tool.ruff.per-file-ignores]
"__init__.py" = ["E402"]
"tests/**" = ["S101", "S105"]

[tool.ruff.mccabe]
max-complexity = 13

[tool.ruff.isort]
known-first-party = ["granian", "tests"]
lines-after-imports = 2
lines-between-types = 1
```

And it can grow from here, by necessity.

pyproject.toml
```diff
@@ -61,5 +63,20 @@ build-backend = "maturin"
...
+
 [tool.pytest.ini_options]
 asyncio_mode = "auto"
+
+[tool.black]
+color = true
+line-length = 120
+target-version = ['py38', 'py39', 'py310', 'py311']
+skip-string-normalization = true
+skip-magic-trailing-comma = true
```
Aeron, line 82: If we don’t want to break the Black’s rules, then it’ll be enough to have line length and target versions:

```suggestion
[tool.black]
line-length = 120
target-version = ['py38', 'py39', 'py310', 'py311']
```

lautjy: 2 cents: The close configs are to `black` defaults the better it is.

gi0baro:
@matthiask @Aeron @lautjy thank you all for your comments and feedbacks.
I should have managed to add the majority of your suggestions; regarding black I really prefer to have single quote strings (you type 'em faster) and thus leave the thing to ruff, while on the pre-commit thing I'd like to think about it in a separated discussion.

I'm merging this right now, as in the next few days I'll be AFK, and at least existing open PRs can solve eventual conflicts. I'm obviously open to further enhancements following this PR :)

lautjy:
Your choice is valid. Just want to add that you could write ' or ", and if you have "format on save" in your IDE, it would run black and automatically convert it to ". So having the default config doesn't force you to change habits - and you will quickly get used to seeing the code (have been thru the change years ago).
IMHO the more repos in default black the better. But at this point we are veering too close to nitpicking. Your repo, your choice :)

PR #140, opened 2023-10-15
Support configuration with environment variables

dekkers:
In a lot of circumstances such as when using containers it easier to do configuration using environment variables than specifying the configuration using command line options. This adds support for configuring Granian using environment variables by settings the Click `auto_envvar_prefix` option.


gi0baro:
@dekkers thank you for your contribution!

PR #159, opened 2023-12-28
WIP: Serve partial files

myers:
- [ ] needs to add header to advertise support for partial files
- [ ] return correct response if the range requested isn't supported
- [ ] return the whole file if the range requested is invalid
- [ ] example should serve whole directory and check symlinks to only serve files in that directory
- [ ] document why this is better than just read/seeking in python code
- [ ] update RSGI spec with start, end parameters

### DIFF
--- a/example_partial_file.py
+++ b/example_partial_file.py
@@ -1,0 +1,35 @@
+import pprint
+
+
+# example of a RSGI handler that can serve partial files
+# `granian --interface rsgi example_partial_file:app`
+# TODO: needs to add header to advertise support for partial files
+# TODO: return correct response if the range requested isn't supported
+# TODO: return the whole file if the range requested is invalid
+# TODO: example should serve whole directory and follow symlinks to make sure 
+# TODO: document why this is better than just read/seeking in python code
+# TODO: update RSGI spec with start, end parameters
+async def app(scope, proto):
+    assert scope.proto == 'http'
+    start = None
+    end = None
+    status = 200
+    if 'range' in scope.headers:
+        http_range = scope.headers.get('range')
+        start_offset, end_offset = http_range.split('=')[1].split('-')
+        start = int(start_offset)
+        if end_offset.isdigit():
+            end = int(end_offset)
+        print("start", start)
+        print("end", end)
+    if start:
+        status = 216
+    proto.response_file(
+        status=status,
+        headers=[
+            ('content-type', 'text/plain'),
+        ],
+        file=__file__,
+        start=start,
+        end=end
+    )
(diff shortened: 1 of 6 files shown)


gi0baro:
@myers thank you for the PR, FYI I'm AFK during these days, I'll look at this after Jan 3rd

gi0baro:
@myers I'm sorry for the late reply, but I finally managed to take a look at this.
I also wanted to release 1.0 before this, as the Hyper API changed and I didn't want to add protocol changes before porting the code to Hyper 1.0, thus this will need changes to adapt the code to 1.0.

That said, I'm not super intrigued by altering `response_file` method with all that logic; also considering the default response code for partial content should be `206`, maybe the best option here is to add an additional method to RSGI spec like `response_file_partial` or `response_file_range`, WDYT?
Also, I would leave the headers management to the application, not the server itself.

myers:
@gi0baro I'm not yet happy with the tests, but here is response_file_partial as requested.

gi0baro (commented):
@myers thank you for your efforts, and sorry for the late review.

I'm quite confused on why we need `granian.utils.range` at all.
To my perspective, `response_file_partial` should behave similarly to `response_file`: in the same way we don't do any specific validation for the latter, I don't think Granian should validate the ranges provided by the application.

I think keeping the server logic as lean as possible in RSGI is the key here. In the same way the RSGI application is responsible to provide the proper headers when responding with a file, it should have the same responsibility also for the ranges.
WDYT?

myers:
> I'm quite confused on why we need granian.utils.range at all.

My other project was building django-rsgi, which builds a native rsgi adaptor for django.  In it I needed to parse the Range header.  It felt like that would be useful for anyone wanting to support that header.   If you don't see the need for it in granian, that's fair.  I think I would break it out into it's own library so that it can be used by others.

You'll note that the `Range` handling is done in python, and that the tests for response_file_partial use get parameters rather than the header to set the response_file_partial args.  This simplified the rust code.

Other than where the code for the Range header should be, what are your thoughts?

gi0baro:
> I needed to parse the Range header. It felt like that would be useful for anyone wanting to support that header. If you don't see the need for it in granian, that's fair. I think I would break it out into it's own library so that it can be used by others.

Yep, I think it's outside the scope of Granian itself.
Maybe we can discuss of a `rsgi` python package: it might make sense to have such utils in something like that.

> Other than where the code for the Range header should be, what are your thoughts?

Other than simplify the code further given we drop all the unnecessary code, I think it's good.

myers:
@gi0baro I've removed the range parsing code.

myers:
w00t!

PR #225, opened 2024-03-03
Add watchfile reload options

Andrew-Chen-Wang:
* Includes reload path option and filter options

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ A Rust HTTP server for Python applications.
 The main reasons behind Granian design are:
 
 - Have a single, correct HTTP implementation, supporting versions 1, 2 (and eventually 3)
-- Provide a single package for several platforms 
+- Provide a single package for several platforms
 - Avoid the usual Gunicorn + uvicorn + http-tools dependency composition on unix systems
 - Provide stable [performance](https://github.com/emmett-framework/granian/blob/master/benchmarks/README.md) when compared to existing alternatives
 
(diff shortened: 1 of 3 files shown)


gi0baro (commented):
@Andrew-Chen-Wang thank you for your contribution. I have a couple of comments on this:
- I would rather use `Path` options in CLI instead of plain strings, given that both `click` and `typer` supports them and also allow additional validation on those (https://typer.tiangolo.com/tutorial/parameter-types/path/#path-validations)
- I'd like to see better descriptions for the new parameters, as IMHO it might be hard for users to understand the difference between `ignore-paths`, `ignore-dirs` and `ignore-entity-patterns`

gi0baro:
> Hi thanks for the response. Will update the help text. In regards to the Path type, because watch file excepts globs, I didn't think Path would work.

From what I can see in `watchfiles` code, the only argument that might not be a path is the entity-patterns one.
Also, I just realised that the CLI parameters for the `ignore-*` should be optional with `None` as defaults, otherwise the check in here https://github.com/samuelcolvin/watchfiles/blob/main/watchfiles/filters.py#L115-L120 will remove defaults having those parameters as empty lists.

Lendemor:
Hello, any update on this PR ? 

I've been testing to replace uvicorn with granian but this options would be needed :)

gi0baro:
@Lendemor I was waiting for @Andrew-Chen-Wang to make the changes we discussed; I don't know if there are any updates from his side.

Given the amount of time this is opened, for sure it would need a rebase from the current master.

Andrew-Chen-Wang:
would rec u just take my changes and make a PR :P not sure when i'll come back

gi0baro:
Closing as superseded by #362

PR #226, opened 2024-03-03
Add Workers and partial ListenerHolder typing

Andrew-Chen-Wang:


granian/_granian.pyi
```diff
@@ -58,3 +59,35 @@ class WSGIScope:
...
+    ):
+        ...
+
+class ASGIWorker(__WorkerConfig):
+    ...
+
+class RSGIWorker(__WorkerConfig):
+    ...
+
+class WSGIWorker(__WorkerConfig):
```
gi0baro, line 87: `WSGIWorker` doesn't accept `websockets_enabled` and `opt_enabled` parameters

granian/_granian.pyi
```diff
@@ -58,3 +59,35 @@ class WSGIScope:
...
+    ...
+
+
+class ListenerHolder:
+    @classmethod
+    def from_address(cls, bind_addr: str, port: int, backlog: int) -> "ListenerHolder": ...
```
gi0baro, line 93: I would add `from __future__ import annotations` and remove the string annotation

gi0baro (changes requested):

granian/_granian.pyi
Andrew-Chen-Wang: Is __future__ annotations in re of Protocol? 

also unless i'm mistaken, the class method doesn't return the instance?

gi0baro: Yup, `ListenerHolder.from_address` returns an instance, see https://github.com/emmett-framework/granian/blob/348d3c71e0e1923688248e2d0744275b7a14129e/src/tcp.rs#L34

I commented about `__future__` so you can rewrite the upper as:

```python
class ListenerHolder:
    @classmethod
    def from_address(cls, bind_addr: str, port: int, backlog: int) -> ListenerHolder: ...
```

gi0baro:
Close as stale

Issue #261, opened 2024-04-02
Changed dependency structure of typer.


WAKayser:
In the 0.12.0 release of typer tiangolo added extra additional dependencies that are installed unless a different package: "typer-slim" is being used.  

I am willing to provide a pull request to either update the dependencies to use the typer-slim package or alternatively I can also update the cli to use the underlying "click" library.

I am wondering what your stance is on this subject?

gi0baro:
@WAKayser thank you for notifying this, I wasn't aware of the changes in Typer.

I'm actually more prone to release a patch version of Granian specifying a typer version < 0.12 to quickly address this, then a PR to drop typer and directly use `click` is more than welcome. I initially enjoyed typer just for the typing system, bit nowadays is become quite bloated in my opinion. Given the scope of Granian CLI, `click` would be definitely enough.

bluetech:
In my opinion, even argparse is good enough, and it brings 0 dependencies :wink:

PR #270, opened 2024-04-13
Rewrite CLI to use click instad of typer

WAKayser:
As discussed in #261 typer changed the scope of the project and started pulling in other dependencies. 

This pull request changes the dependency used from typer to click. 
The resulting code is a bit more verbose, but the general way of working with it is largely similar. 
I found that there were two features missing in click.

First is no support for enums as choices. But this is easy to work around by supplying a list of choices and in the function itself convert to an enum. 
Second feature missing was the option to show the environment variables for each option. I have added a simple monkey patch to the option function to extend the help text. 

Current state is functional, all option types work and environment variables in general work.. But some more testing is wise.

### DIFF
--- a/granian/cli.py
+++ b/granian/cli.py
@@ -37,7 +37,7 @@ def pretty_print_default(value: Optional[bool]) -> Optional[str]:
     return value
 
 
-def show_env_var(*param_decls: str, cls: Optional[Type[click.Option]] = None, **attrs: Any) -> Callable[[FC], FC]:
+def documented_option(*param_decls: str, cls: Optional[Type[click.Option]] = None, **attrs: Any) -> Callable[[FC], FC]:
     """This is the option generator from Click.
 
     It has a very simple update so that the environment variable will also get added to the help text.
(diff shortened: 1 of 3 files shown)


gi0baro (commented):
This looks good in general.

I would really love to re-use the existing enums though, just to avoid PITAs in the future if those enums gets changed.

A simple function like this should be enough:

```python
def enum_to_click_choice(v):
    return click.Choice([member.value for member in v])
```

so we can write 

```python
type=enum_to_click_choice(Interfaces)
```

granian/cli.py
```diff
@@ -1,135 +1,254 @@
...
+            attrs['show_default'] = pretty_print_default(attrs['default'])
+        # end change
+        click.decorators._param_memo(f, cls(param_decls, **attrs))
+        return f
+
+    return decorator
+
+
+# monkey patch so that the behaviour is more similar to the docs.
+click.option = show_env_var
```
gi0baro, line 62: I just thought monkey patching `click.option` might not be a good idea, as several frameworks/apps used with Granian might depend on it.

Maybe we should just update `show_env_var` to be a `cli_option` we use in place of `click.option` that will just return `click.option` decorator patching the arguments in advance.

WAKayser: That's true. Updated the PR.

WAKayser:
I have added a simple EnumType to create the click.Choice from the enum. This is based on: https://github.com/pallets/click/issues/605#issuecomment-917022108

Finally I have also added a pretty printer to print the defaults a bit neater. This allows to turn all booleans into "disabled" or "enabled" without duplicate configurations.

Finally, I have also finished testing. This was primarily done by adding the following snippet to the start of the cli function.  Then check the diff between the version from this PR and what is currently on pypi. 
```python
    print(f'{app=}, {host=}, {port=}, {interface=}, {http=}, {websockets=}, {workers=}, {threads=}, {blocking_threads=}')
    print(f'{loop=}, {loop_opt=}, {backlog=}, {http1_buffer_size=}, {http1_keep_alive=}, {http1_pipeline_flush}')
    print(f'{http2_adaptive_window=}, {http2_initial_connection_window_size=}, {http2_initial_stream_window_size}')
    print(f'{http2_keep_alive_interval=}, {http2_keep_alive_timeout=}, {http2_max_concurrent_streams=}')
    print(f'{http2_max_frame_size=}, {http2_max_headers_size=}, {http2_max_send_buffer_size=}, {log_enabled=}')
    print(f'{log_level=}, {log_config=}, {ssl_keyfile=}, {ssl_certificate=}, {url_path_prefix=}')
    print(f'{respawn_failed_workers=}, {reload=}, {process_name=}')
    return
```
This returned the same when using all the defaults and with all values set via either the command options or via environment variables.

PR #274, opened 2024-04-16
Add --respawn-interval option

elbaro:
Add --reload-interval [seconds] so that Granian can reload slowly for large applications.

I could not locally test this PR


gi0baro:
Rebased and renamed to `--respawn-interval`

PR #277, opened 2024-04-22
Improve typing for python code

TheJecksMan:
This is my first PR ever😊.

This pull request **improves and fixes** the typing for some modules to avoid future bugs (Some methods were called without implicit type annotation)

I am willing to create a discussion if needed.


granian/_types.py
```diff
@@ -1,6 +1,24 @@
-from typing import Union
+from typing import Optional, Union
 
 
 class WebsocketMessage:
     kind: int
     data: Union[bytes, str]
+
+
+class HTTP1ParamType:
```
gi0baro, line 9: Why do we need this, given the existence of dataclasses in https://github.com/emmett-framework/granian/blob/master/granian/http.py?

granian/_loops.py
```diff
@@ -8,9 +8,9 @@
 
 
 class Registry:
-    __slots__ = ['_data']
+    __slots__ = ('_data',)
 
-    def __init__(self):
+    def __init__(self) -> None:
```
gi0baro, line 13: Isn't `-> None` implicit? Do we really need to bloat code with such useless annotations?

granian/_loops.py
```diff
@@ -19,7 +19,7 @@ def __contains__(self, key: str) -> bool:
     def keys(self) -> Iterable[str]:
         return self._data.keys()
 
-    def register(self, key: str) -> Callable[[], Callable[..., Any]]:
+    def register(self, key: str) -> Callable[[Callable[..., Any]], Callable[..., Any]]:
```
gi0baro, line 22: Probably would be worth to define a custom type `Wrappable = Callable[..., Any]` and cleanup the code a bit

granian/_loops.py
```diff
@@ -34,12 +34,14 @@ def get(self, key: str) -> Callable[..., Any]:
 
 
 class BuilderRegistry(Registry):
-    __slots__ = []
+    __slots__ = ('_data',)
```
gi0baro, line 37: `_data` slot is already inherited by the upper class. An empty list (and this is also why I used lists in place of tuples, makes the thing more readable IMHO) just avoid to add the local dictionary and should be fine.

granian/_loops.py
```diff
@@ -34,12 +34,14 @@ def get(self, key: str) -> Callable[..., Any]:
...
 
-    def __init__(self):
-        self._data: Dict[str, Tuple[Callable[..., Any], List[str]]] = {}
+    def __init__(self) -> None:
```
gi0baro, line 39: Same as before

granian/wsgi.py
```diff
@@ -1,13 +1,13 @@
...
-from typing import Any, List, Tuple
+from typing import Any, Callable, Dict, List, Tuple
 
 
 class Response:
-    __slots__ = ['status', 'headers']
+    __slots__ = ('status', 'headers')
 
-    def __init__(self):
+    def __init__(self) -> None:
```
gi0baro, line 10: Same as the previous comments

granian/wsgi.py
```diff
@@ -16,10 +16,10 @@ def __call__(self, status: str, headers: List[Tuple[str, str]], exc_info: Any =
         self.headers = headers
 
 
-def _callback_wrapper(callback, scope_opts):
+def _callback_wrapper(callback: Callable[..., Any], scope_opts: Dict[str, Any]):
     basic_env = dict(os.environ)
     basic_env.update(
-        {
+        **{
```
gi0baro, line 22: Why adding the `**` operator here?

gi0baro (commented):
@TheJecksMan thank you for your contribution.

I left some specific comments on your code, feel free to discuss those further.
While I appreciate the fixes to existing typing annotations, I'm not a huge fan on those, especially on private/internal methods and interfaces, given such objects would never be exposed to 3rd party, and thus the point of having annotations seems a bit unworthy to me.

For instance, instead of type annotations on the default `_spawn_*` methods, would probably be more valuable for developers to add a `SpawnMethod` type which maps to a `Callable` with the arguments you typed on `_spawn` methods and add such annotation on the `spawn_target` argument of `Granian.serve`.
Or, as another example, given the only method used on `ListenerHolder` is `from_address` (`__getstate__` is a generally-available magic method), I would have personally annotated only that one.

Let me know WDYT :)

granian/wsgi.py
TheJecksMan: Good point @gi0baro. I think you are right, unpacking the dictionary is unnecessary. I did it just to fix the **pylance** bug, but I think here it is enough to explicitly specify the type for the solution. This will be fixed.

granian/_loops.py
TheJecksMan: Then I would prefer to use `WrappableT = TypeVar("WrappableT", bound=Callable[..., Any])` for more flexibility and type constraint. What is your opinion?

gi0baro: 👍 makes sense

TheJecksMan:
The code has been corrected. But I don't quite understand the idea with `SpawnMethod`. If you need a type that annotated all incoming parameters to `_spawn_*` methods, wouldn't that be too overhead? But then again, maybe I'm misunderstanding.

Issue #276, opened 2024-04-17
Possible memory-leak analysis


gi0baro:
Currently suspected protocol: ASGI.
Verification on RSGI and WSGI should be done as well.

### Discussed in https://github.com/emmett-framework/granian/discussions/268

<div type='discussions-op-text'>

<sup>Originally posted by **cpoppema** April 12, 2024</sup>
For two projects I replaced uvicorn with uvloop with granian and my monitoring tools show an increase in memory usage. I am wondering if this is something I have to accept with granian - I don't know if it'll flatten out at some point - or if there is an underlying issue I might be able to help debug. Benchmarking is all about request numbers but does not mention any memory increase, so I suspect an issue.

Figures below are from 2 services running fastapi + pydantic + sqlalchemy.

service 1, graph is from 7 days:

cpu before (uvicorn)/after (granian)
![2024-04-12-1712919886_screenshot](https://github.com/emmett-framework/granian/assets/1432901/dcf5376d-47ad-4c82-8c84-4bcb7b42a6b7)

requests before (uvicorn)/after (granian)
![2024-04-12-1712919843_screenshot](https://github.com/emmett-framework/granian/assets/1432901/d38b2644-5e5b-4ed1-bc22-922230abae8d)

memory before (uvicorn)/after (granian)
![2024-04-12-1712919806_screenshot](https://github.com/emmett-framework/granian/assets/1432901/6a9f56a7-5144-482b-af95-5ccecc491520)

service 2, graph is from 2 days:

cpu before (uvicorn)/after (granian)
![2024-04-12-1712920252_screenshot](https://github.com/emmett-framework/granian/assets/1432901/e2dfed36-a4f8-49a0-bdad-e20a4fb9c018)

requests before (uvicorn)/after (granian)
![2024-04-12-1712920168_screenshot](https://github.com/emmett-framework/granian/assets/1432901/585c2933-5880-4627-a70c-7249829748fc)

memory before (uvicorn)/after (granian)
![2024-04-12-1712920305_screenshot](https://github.com/emmett-framework/granian/assets/1432901/29697785-c24e-45ce-887e-15d64d437ee0)

</div>

gi0baro:
@cpoppema jfyi I did some investigation on this during the weekend using https://github.com/bloomberg/memray, but I wasn't able to reproduce the memory increase trend you had on my machine (macOS). I need to schedule a similar test on a Linux machine, as it might be a target-specific issue.

cpoppema:
[...]

PR #279, opened 2024-04-25
Set error result on awaitables when cancelling futures

### LINKED ISSUE #252: `1.2.0` doesn't shut down properly
When upgrading to `v1.2.0` it seems the shutdown for `asgi` doesn't work properly - it seems to be hanging.

sciyoshi:
This seems to fix #252 and possibly #276 as well.

### DIFF
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -290,7 +290,7 @@ where
     rt.spawn(async move {
         tokio::select! {
             result = fut => aw.get().set_result(result, aw_ref),
-            () = cancel_tx.notified() => {}
+            () = cancel_tx.notified() => aw.get().set_result(Err::<T, pyo3::PyErr>(pyo3::exceptions::asyncio::CancelledError::new_err("Task cancelled")), aw_ref)
         }
     });
 


sciyoshi:
Nevermind, I still see a shutdown issue happening with my main test case, although this fixes the simple repro provided in https://github.com/emmett-framework/granian/issues/252#issuecomment-2064512425.

gi0baro:
@sciyoshi I think this doesn't solve anything: probably it seems to fix #252 just because the overall behaviour is flaky.
In fact, I'm quite sure this is actually wrong, as the cancellation output is already covered by https://github.com/emmett-framework/granian/blob/master/src/callbacks.rs#L224 and https://github.com/emmett-framework/granian/blob/master/src/callbacks.rs#L237, which is set on the internal state directly by Python here https://github.com/emmett-framework/granian/blob/master/src/callbacks.rs#L202

gi0baro:
@sciyoshi _maybe_ the actual issue might be with this line https://github.com/emmett-framework/granian/blob/master/src/callbacks.rs#L120, as it might overwrite the cancelled state.
Probably a more error-prone strategy would be to check the state value before setting the completed one.

sciyoshi:
Fair point, it might be that the state isn't getting set correctly.

PR #308, opened 2024-05-27
Additional access log variables

nitrag:
- Response (Body) Length
- User Agent, if available

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,6 +218,8 @@ The following atoms are available for use:
 | method | Request HTTP method |
 | scheme | Request scheme |
 | protocol | HTTP protocol version |
+| response\_length | Size (in bytes) of the HTTP response |
+| user\_agent | User Agent of the HTTP request, if available |
 
 ### Processes and threads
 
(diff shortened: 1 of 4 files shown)


gi0baro (changes requested):
Thank you for your contribution, but the provided code is both wrong and lacking what you described in the descriptions, specifically:
- missing RSGI protocol implementation
- `scope['CONTENT_LENGTH']` in WSGI is the request size, not the response one
- ASGI won't have the provided keys in scope

I can't accept this given the upper points.

nitrag:
Unfortunately, I'm new to Rust. This repository was either crafted by black magic or written by a genius. I have no idea what I'm doing. Having these features would bring it to feature parity with gunicorn so it can be replaced.

gi0baro:
> Unfortunately, I'm new to Rust. This repository was either crafted by black magic or written by a genius. I have no idea what I'm doing. Having these features would bring it to feature parity with gunicorn so it can be replaced.

Please open up an issue to request the change, eventually someone else will address it.
Closing this in the meantime.

PR #327, opened 2024-06-13
Implemented `--pid-file` to write the process id to a file

### LINKED ISSUE #319: Implement --pid
Hi guys

Thanks for having created granian, it's a great thing!

There's of course always potential for improvement, here's something I miss from gunicorn:

- In gunicorn you can pass [--pid](https://docs.gunicorn.org/en/stable/settings.html#pidfile) which will write the gunicorn's pid (not the workers) into a simple file
- Using that PID you can then send signals to the Main process, see https://docs.gunicorn.org/en/stable/signals.html
Most important one is probably `HUP` which will do a restart of the workers, which is very nice

Not sure I find the time to implement this, especially as I'm not a UNIX pro, but I could give it a try if no one wants to do this

stefins:
Hi,

This PR closes #319

Thanks :)

### DIFF
--- a/granian/cli.py
+++ b/granian/cli.py
@@ -6,7 +6,7 @@ from typing import Any, Callable, Optional, Type, TypeVar, Union
 import click
 
 from .constants import HTTPModes, Interfaces, Loops, ThreadModes
-from .errors import ConfigurationError
+from .errors import ConfigurationError, PidFileError
 from .http import HTTP1Settings, HTTP2Settings
 from .log import LogLevels
 from .server import Granian
@@ -198,6 +198,11 @@ def option(*param_decls: str, cls: Optional[Type[click.Option]] = None, **attrs:
     '--process-name',
     help='Set a custom name for processes (requires granian[pname] extra)',
 )
+@option(
+    '--pid-file',
+    type=click.Path(exists=False, file_okay=True, dir_okay=False, writable=True, path_type=pathlib.Path),
+    help='A filename to use for the PID file.',
+)
 @click.version_option(message='%(prog)s %(version)s')
 def cli(
     app: str,
@@ -238,6 +243,7 @@ def cli(
     respawn_interval: float,
     reload: bool,
     process_name: Optional[str],
+    pid_file: Optional[pathlib.Path],
 ) -> None:
     log_dictconfig = None
     if log_config:
@@ -289,11 +295,15 @@ def cli(
         respawn_interval=respawn_interval,
         reload=reload,
         process_name=process_name,
+        pid_file=pid_file,
     )
 
     try:
         server.serve()
     except ConfigurationError:
+        raise click.exceptions.Exit(1)
+    except PidFileError as err:
+        print(err)
         raise click.exceptions.Exit(1)
 
 
--- a/granian/cli.py
+++ b/granian/cli.py
@@ -6,7 +6,7 @@ from typing import Any, Callable, Optional, Type, TypeVar, Union
 import click
 
 from .constants import HTTPModes, Interfaces, Loops, ThreadModes
-from .errors import ConfigurationError, PidFileError
+from .errors import FatalError
 from .http import HTTP1Settings, HTTP2Settings
 from .log import LogLevels
 from .server import Granian
@@ -201,7 +201,7 @@ def option(*param_decls: str, cls: Optional[Type[click.Option]] = None, **attrs:
 @option(
     '--pid-file',
     type=click.Path(exists=False, file_okay=True, dir_okay=False, writable=True, path_type=pathlib.Path),
-    help='A filename to use for the PID file.',
+    help='A path to write the PID file to',
 )
 @click.version_option(message='%(prog)s %(version)s')
 def cli(
@@ -300,10 +300,7 @@ def cli(
 
     try:
         server.serve()
-    except ConfigurationError:
-        raise click.exceptions.Exit(1)
-    except PidFileError as err:
-        print(err)
+    except FatalError:
         raise click.exceptions.Exit(1)
 
 
--- a/granian/server.py
+++ b/granian/server.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import contextvars
+import errno
 import multiprocessing
 import os
 import signal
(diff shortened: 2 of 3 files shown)


granian/cli.py
```diff
@@ -198,6 +198,11 @@ def option(*param_decls: str, cls: Optional[Type[click.Option]] = None, **attrs:
     '--process-name',
     help='Set a custom name for processes (requires granian[pname] extra)',
 )
+@option(
+    '--pid',
```
gi0baro, line 202: Maybe a little of a *nitpick* here, but I would name the parameter `--pid-file`

granian/cli.py
```diff
@@ -198,6 +198,11 @@ def option(*param_decls: str, cls: Optional[Type[click.Option]] = None, **attrs:
...
+    type=click.Path(exists=False, file_okay=True, dir_okay=False, readable=True, path_type=pathlib.Path),
```
gi0baro, line 203:
```suggestion
    type=click.Path(exists=False, file_okay=True, dir_okay=False, writable=True, path_type=pathlib.Path),
```

granian/server.py
```diff
@@ -95,6 +96,7 @@ def __init__(
         respawn_interval: float = 3.5,
         reload: bool = False,
         process_name: Optional[str] = None,
+        pid: Optional[Path] = None,
```
gi0baro, line 99: Same for the CLI parameter, I'd change this to `pid_file`

granian/server.py
```diff
@@ -461,9 +465,50 @@ def setup_signals(self):
...
+                    self._write_pid()
+                    return
+            try:
+                os.kill(old_pid, 0)
+            except OSError as e:
+                if e.args[0] == errno.ESRCH:
+                    self._write_pid()
+                    return
+                else:
+                    raise e
```
gi0baro, line 490: This looks a bit too overcomplicated to me..
I'd say Granian shouldn't try to kill other processes; checking if the file exists before writing and raising an error in such case is more than enough IMHO.

granian/server.py
```diff
@@ -461,9 +465,50 @@ def setup_signals(self):
...
+        self._write_pid()
+
+    def _unlink_pidfile(self):
+        if not (self.pid and self.pid.exists()):
+            return
+        with self.pid.open('r') as pid_file:
+            file_pid = int(pid_file.read())
+
+        if file_pid == self.main_pid:
+            self.pid.unlink()
```
gi0baro, line 506: Same for writing the pid file: I think Granian should just try to unlink the file if it exists, without additional checks.

granian/server.py
```diff
@@ -461,9 +465,50 @@ def setup_signals(self):
...
+                if e.args[0] == errno.ESRCH:
+                    self._write_pid()
+                    return
+                else:
+                    raise e
+            if old_pid == self.main_pid:
+                return
+            msg = "Already running on PID %s (or pid file '%s' is stale)"
+            logger.error(msg % (old_pid, self.pid))
+            sys.exit(1)
```
gi0baro, line 495: This should probably be a custom `Exception` that we catch in `cli.py` and exit there, in the same way we do for `ConfigurationError`.

granian/server.py
stefins: `os.kill(old_pid, 0)` is not trying to kill the particular process. Its just trying to find if the particular pid exists in system, If not we just overwrite the `pid_file` with the new pid .

But If you don't want to handle this case. I'll be happy to remove that change from this PR..

granian/server.py
stefins: If I unlink a file that already exists with a different pid / content . that could potentially remove the file with unrelated changes than that of `granian`'s `main_pid`.

Again same, I can remove the additional checks if that look like an overkill.

granian/server.py
gi0baro: Understood.
Let's keep it, but change line 481 https://github.com/emmett-framework/granian/pull/327/files/405ff32e5fbe32a1ac6db956e6d9df2c71913be8#diff-571aaa34f48968993dca3f59fcc89aff6e892451c088006d134ab11c0acf0cc2R481 to be more conservative: if we cannot read the pid from that file, let's just stop and raise, 'cause probably it's not written by Granian.

granian/server.py
gi0baro: Ok, let's keep it but wrap the file read in a `try` block just to avoid unmanaged errors on server shutdowns.

granian/server.py
stefins: Addressed via https://github.com/emmett-framework/granian/compare/405ff32e5fbe32a1ac6db956e6d9df2c71913be8..ea4efa7dad1e4b1fbe09d2dfd822936be9d2095d

granian/server.py
stefins: Addressed via https://github.com/emmett-framework/granian/compare/405ff32e5fbe32a1ac6db956e6d9df2c71913be8..ea4efa7dad1e4b1fbe09d2dfd822936be9d2095d

granian/server.py
stefins: Addressed via https://github.com/emmett-framework/granian/compare/405ff32e5fbe32a1ac6db956e6d9df2c71913be8..ea4efa7dad1e4b1fbe09d2dfd822936be9d2095d

granian/server.py
stefins: Addressed via https://github.com/emmett-framework/granian/compare/405ff32e5fbe32a1ac6db956e6d9df2c71913be8..ea4efa7dad1e4b1fbe09d2dfd822936be9d2095d

Let me know if the name of the custom `Exception` need to be changed.

granian/cli.py
stefins: Addressed via https://github.com/emmett-framework/granian/compare/405ff32e5fbe32a1ac6db956e6d9df2c71913be8..ea4efa7dad1e4b1fbe09d2dfd822936be9d2095d

granian/cli.py
stefins: Addressed via https://github.com/emmett-framework/granian/compare/405ff32e5fbe32a1ac6db956e6d9df2c71913be8..ea4efa7dad1e4b1fbe09d2dfd822936be9d2095d

gi0baro:
@stefins pushed a refactor commit https://github.com/emmett-framework/granian/pull/327/commits/de3134ae98c069a155ba97ddc6595b93407059af

If you don't have any objections to my commit, I'll merge this.

stefins:
@gi0baro I just tested this locally and when I try to run a second instance of `granian` with the same `--pid-file` It's overwriting the pid of the first process. Is that the expected behavior?

imo We shouldn't overwrite the `pid-file` if another `granian` instance is already using that particular `pid-file`

gi0baro:
> @gi0baro I just tested this locally and when I try to run a second instance of `granian` with the same `--pid-file` It's overwriting the pid of the first process. Is that the expected behavior?
> 
> imo We shouldn't overwrite the `pid-file` if another `granian` instance is already using that particular `pid-file`

My bad, should be fixed with https://github.com/emmett-framework/granian/pull/327/commits/be44b9722524a2c64dd92b0d63f6dfb34e0d0d66

PR #362, opened 2024-07-31
feat: better reload options

### LINKED ISSUE #353: Add Granian extended reloading options
It would be nice to define included or excluded dirs (or paths)  for reloading

## Funding
* You can sponsor this specific effort via a [Polar.sh](https://polar.sh) pledge below
* We receive the pledge once the issue is completed & verified

<a href="https://polar.sh/emmett-framework/granian/issues/353">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/emmett-framework/granian/issues/353/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

iamkhav:
Hey all,
I based this PR off of @Andrew-Chen-Wang's work and included his original commit from #225.

Additionally to the rebase, I made a few changes to the descriptions (as required in the previous PR) and added the capability to pass a `reload_filter` from the Python API.

The `reload_filter` should be an obj based on watchfiles' `BaseClass`.

Closes #353 and #225.


granian/server.py
```diff
@@ -95,6 +96,11 @@ def __init__(
         respawn_failed_workers: bool = False,
         respawn_interval: float = 3.5,
         reload: bool = False,
+        reload_paths: Sequence[pathlib.Path] = (pathlib.Path.cwd(),),
+        reload_ignore_dirs: Optional[Sequence[str]] = (),
+        reload_ignore_entity_patterns: Optional[Sequence[str]] = (),
+        reload_ignore_paths: Optional[Sequence[pathlib.Path]] = (),
+        reload_filter: Optional[BaseFilter] = None,
```
iamkhav, line 103: Actually we could (or should) limit this to objs of subclasses only 🤔

The docstring of `BaseFilter` states
> Useful base class for creating filters. `BaseFilter` should be inherited and configured, rather than used directly.

so it might make more sense to use
```suggestion
        reload_filter: Optional[Type[BaseFilter]] = None,
```

gi0baro: Makes sense 👍

granian/cli.py
```diff
@@ -194,6 +194,38 @@ def option(*param_decls: str, cls: Optional[Type[click.Option]] = None, **attrs:
...
+    '--reload-ignore-dirs',
+    default=(),
+    help=(
+        'Names of directories to ignore (i.e. should not trigger reload). '
+        'Extends the default list of directories to ignore in watchfiles.filters.DefaultFilter.'
+    ),
+    multiple=True,
+)
+@option(
+    '--reload-ignore-entity-patterns',
```
gi0baro, line 214: I would rename this to `--reload-ignore-patterns`, just for the sake of simplicity

granian/cli.py
```diff
@@ -242,6 +274,10 @@ def cli(
     respawn_failed_workers: bool,
     respawn_interval: float,
     reload: bool,
+    reload_paths: List[pathlib.Path],
```
gi0baro, line 277: Maybe this should be an `Optional[...]` as well, we can customise the display of the default value in the CLI.

granian/cli.py
```diff
@@ -194,6 +194,38 @@ def option(*param_decls: str, cls: Optional[Type[click.Option]] = None, **attrs:
...
+@option(
+    '--reload-paths',
+    type=click.Path(exists=True, file_okay=True, dir_okay=True, readable=True, path_type=pathlib.Path),
+    default=(pathlib.Path.cwd(),),
+    help='Paths to watch for changes',
+    multiple=True,
+)
+@option(
+    '--reload-ignore-dirs',
+    default=(),
```
gi0baro, line 206: If the option has no default value, I would just skip the whole `default=` thing, just to be consistent with the rest of the code.

granian/server.py
```diff
@@ -95,6 +96,11 @@ def __init__(
         respawn_failed_workers: bool = False,
         respawn_interval: float = 3.5,
         reload: bool = False,
+        reload_paths: Sequence[pathlib.Path] = (pathlib.Path.cwd(),),
+        reload_ignore_dirs: Optional[Sequence[str]] = (),
+        reload_ignore_entity_patterns: Optional[Sequence[str]] = (),
+        reload_ignore_paths: Optional[Sequence[pathlib.Path]] = (),
```
gi0baro, line 102: I would rather make all of these parameters `Optional[...] = None` and set the default later on the `self.var = ` step.

granian/server.py
```diff
@@ -577,11 +588,28 @@ def _serve_with_reloader(self, spawn_target, target_loader):
...
+        ignore_dirs = (*reload_filter.ignore_dirs, *self.reload_ignore_dirs)
+        ignore_entity_patterns = (
+            *reload_filter.ignore_entity_patterns,
+            *self.reload_ignore_entity_patterns,
+        )
+        ignore_paths = (*reload_filter.ignore_paths, *self.reload_ignore_paths)
+        # Construct new filter
+        reload_filter = watchfiles.filters.DefaultFilter(
+            ignore_dirs=ignore_dirs, ignore_entity_patterns=ignore_entity_patterns, ignore_paths=ignore_paths
+        )
```
gi0baro, line 603: I get from this code that there's no way to actually *override* `watchfiles` defaults from the CLI, right?

Given my suggestion on using `None` as default for all these variables, maybe we can use the `DefaultFilter` values only if `self` items are not `None`, wdyt?

gi0baro (commented):
Left some comments; vast majority of them are nits.

granian/server.py
```diff
@@ -4,6 +4,7 @@
 import errno
 import multiprocessing
 import os
+import pathlib
```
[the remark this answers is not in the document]
iamkhav: Ah, inattention on my side.

granian/cli.py
iamkhav: Renamed everywhere, internally as well.

granian/server.py
iamkhav: We can't (comfortably) override an existing `*Filter` obj because of the preprocessing that happens in the `__init__`.
We could override the public members, call the `__init__` again or create a new obj like I did.
Calling the `__init__` again might be prettier, to be honest.

We should probably first think about what the expected behaviour from CLI is. 
1. Extending defaults
2. Replacing defaults

I thought extending the `DefaultFilter` makes sense because normally, one wouldn't want to miss out on the defaults they provide.
https://github.com/samuelcolvin/watchfiles/blob/main/watchfiles/filters.py#L75-L99

If someone **really** wants to replace the `DefaultFilter` defaults, they can use Granian from Python and provide an obj to `reload_filter`.

granian/server.py
iamkhav: Done!

granian/cli.py
iamkhav: Done!

iamkhav:
I'll have to make another manual testing round later on, haven't spent a lot of time on that after the last two commits.

@gi0baro, should we speak about unit tests for some functionalities? I could cover reload options in an early iteration.
I sent you an email with contact info a few days ago.

granian/server.py
gi0baro: > If someone really wants to replace the DefaultFilter defaults, they can use Granian from Python and provide an obj to reload_filter.

gi0baro:
> I'll have to make another manual testing round later on, haven't spent a lot of time on that after the last two commits.

Gonna make another code review ASAP.

> @gi0baro, should we speak about unit tests for some functionalities? I could cover reload options in an early iteration. 

If you figure out a smart way to unit-test reload options, then yes please, go ahead. In general it's hard to unit-test Granian for everything which regards actual HTTP requests/application interaction (given also how the codebase is structured right now), maybe with some *huge refactoring* we can change this, but at the moment it feels too costly for being worth (especially considering there are 0 contributors on the Rust side at the moment except for myself). But again, given the context, I think it is a good candidate for unit.

> I sent you an email with contact info a few days ago.

Yep, I see it. I just need to find some time to proper reply :)

granian/_imports.py
```diff
@@ -5,5 +5,7 @@
 
 try:
     import watchfiles
+    from watchfiles import BaseFilter
```
gi0baro, line 8: maybe we can use `watchfiles.BaseFilter` in *server.py* and leave this untouched? wdyt?

granian/cli.py
```diff
@@ -194,6 +194,35 @@ def option(*param_decls: str, cls: Optional[Type[click.Option]] = None, **attrs:
     default=False,
     help="Enable auto reload on application's files changes (requires granian[reload] extra)",
 )
+@option(
+    '--reload-paths',
+    type=click.Path(exists=True, file_okay=True, dir_okay=True, readable=True, path_type=pathlib.Path),
+    help='Paths to watch for changes',
+    show_default='pathlib.Path.cwd()',
```
gi0baro, line 201:
```suggestion
    show_default='Working directory',
```

granian/cli.py
```diff
@@ -242,6 +271,10 @@ def cli(
     respawn_failed_workers: bool,
     respawn_interval: float,
     reload: bool,
+    reload_paths: List[pathlib.Path],
```
gi0baro, line 274:
```suggestion
    reload_paths: Optional[List[pathlib.Path]],
```

granian/server.py
```diff
@@ -129,6 +134,11 @@ def __init__(
         self.respawn_failed_workers = respawn_failed_workers
         self.reload_on_changes = reload
         self.respawn_interval = respawn_interval
+        self.reload_paths = reload_paths or Path.cwd()
```
gi0baro, line 137:
```suggestion
        self.reload_paths = reload_paths or [Path.cwd()]
```

gi0baro (commented):
Left some suggestions/comments.
Also, probably changes to `README.md` need an update too.

granian/server.py
iamkhav: Arghh sorry, last commit was hastily done.

granian/_imports.py
iamkhav: Ah, I wasn't aware we could type hint to a potential `None.BaseFilter` but it seems to work.

iamkhav:
I'm sorry for the delay on this PR @gi0baro!

Meanwhile, I was thinking a bit about tests.
1. Launch Granian instance with certain reload params
2. Confirm default behaviour for a route (optional)
3. Replace a source file to make that route respond with a specific body
4. Send a request to Granian instance and check if app successfully reloaded by asserting the specific modified body

<sub> E.g. 2. could be an assert for the response body **not** to be the later modified one.

I think we could pull that off with minimal changes to `tests/apps`.
Happy to attempt this either here or in a subsequent PR (but probably before 1.6).

WDYT?

gi0baro:
> I'm sorry for the delay on this PR @gi0baro!

np at all, there's no need to rush here :)

> Meanwhile, I was thinking a bit about tests.
> 
> 1. Launch Granian instance with certain reload params
> 2. Confirm default behaviour for a route (optional)
> 3. Replace a source file to make that route respond with a specific body
> 4. Send a request to Granian instance and check if app successfully reloaded by asserting the specific modified body
> 
> E.g. 2. could be an assert for the response body **not** to be the later modified one.

This sounds more like testing `watchfiles` itself rather than Granian to me.
I would probably approach this with a different strategy, unit-testing specifically:
- the final merged params are indeed what we expect in `Granian` instance
- the change trigger works
I think we can implement those tests without the need of making a request/response flow, as refactoring `Granian` class to mock the startup and the reload trigger should be relatively easy.

> Happy to attempt this either here or in a subsequent PR (but probably before 1.6).

I'm definitely more up to a separate PR for tests, planning to review and merge this one in the near future.
I don't have any specific ETA for 1.6, so feel free to take your time; we can discuss the single PRs individually for 1.6 inclusion :)

granian/cli.py
```diff
@@ -204,16 +204,16 @@ def option(*param_decls: str, cls: Optional[Type[click.Option]] = None, **attrs:
...
+        "Extends the default list of directories to ignore in watchfiles' default filter"
     ),
     multiple=True,
 )
 @option(
     '--reload-ignore-patterns',
     help=(
-        'Entity patterns (regex) to ignore changes for. '
-        'Extends the default list of patterns to ignore in watchfiles.filters.DefaultFilter.'
+        'Path patterns (regex) to ignore changes for. '
```
iamkhav, line 215: Quick note because I stumbled over this while testing the feature.

I had also assumed that this will regex match on the full path of something that triggered the reload. However, this only matches the last part of the path, meaning the exact file/dir that triggered the reload, not the entire path.

https://watchfiles.helpmanual.io/api/filters/#watchfiles.DefaultFilter.ignore_entity_patterns

"Entity" isn't a much better description 😛 but "Path" might be suggesting a different behaviour altogether. WDYT?

<sub>I want to avoid issues where users report this to be broken.

gi0baro: Oh, I missed that. Would *Relative path patterns* make sense to you?

iamkhav: Sorry for the late answer.

I think being fully explicit could help.
```suggestion
        'File/directory name patterns (regex) to ignore changes for. '
```
WDYT?

gi0baro: Looks perfect. Can you open a PR for this? otherwise I'll do it by myself later

iamkhav: No worries, will do!

PR #363, opened 2024-07-31
chore: add `.idea` folder to `.gitignore`

iamkhav:
This will prove useful to us Jetbrains users :)


gi0baro (changes requested):
Can you please move this at line 5 (so we keep logical + alphabetical ordering)?

gi0baro (approved):

PR #369, opened 2024-08-01
feat: more detailed reload logging

iamkhav:
### Description
Hey,
this PR will add slightly more detail to logging on reloads.

Instead of just logging out
```
[INFO] Changes detected, reloading workers..
```
we would give one line for each change like
```
...
[INFO] Started worker-1
[INFO] Started worker-1 runtime-1
[INFO] Changes detected, reloading workers..
[INFO] Modified: /Users/AmonKhavari/granian/a_cool_file.py
[INFO] Modified: /Users/AmonKhavari/granian/another_cool_one.py
[INFO] Added: /Users/AmonKhavari/granian/something
[INFO] Deleted: /Users/AmonKhavari/granian/something_else
[INFO] Stopping worker-1
[INFO] Stopping worker-1 runtime-1
...
```

I personally think it will contribute to better visibility on potential misconfiguration that a user might face.
Especially if #362 goes to master and things might need some configuration.

However, I realize this is fairly verbose. I figured it's okay since it only logs out on `INFO` level.

Let me know what you think :)

### DIFF
--- a/granian/server.py
+++ b/granian/server.py
@@ -581,8 +581,10 @@ class Granian:
         sock = self.startup(spawn_target, target_loader)
 
         try:
-            for _ in watchfiles.watch(reload_path, stop_event=self.main_loop_interrupt):
+            for changes in watchfiles.watch(reload_path, stop_event=self.main_loop_interrupt):
                 logger.info('Changes detected, reloading workers..')
+                for change, file in changes:
+                    logger.info(f'{change.raw_str().capitalize()}: {file}')
                 self._stop_workers()
                 self._spawn_workers(sock, spawn_target, target_loader)
         except StopIteration:


gi0baro (commented):
The only possible issue I see for this is to bloat the logs with a lot of information in some cases (eg: changing branch with `git`); but at the same time the reloader shouldn't be used in production environments, so probably not a real issue anyways.

Maybe we can use `logger.debug()` for the changelist, so it gets printed only with that verbosity level (and thus supposedly during development), wdyt?

iamkhav:
> The only possible issue I see for this is to bloat the logs with a lot of information in some cases (eg: changing branch with `git`); but at the same time the reloader shouldn't be used in production environments, so probably not a real issue anyways.

I agree, reload will most likely only happen on dev machines. 

> Maybe we can use `logger.debug()` for the changelist, so it gets printed only with that verbosity level (and thus supposedly during development), wdyt?

I've got a mixed relationship with `logger.debug()`, some packages absolutely spam it (e.g. boto3) and if a user goes the *very short way* of only giving the root logger the desired loglevel, they might be absolutely flooded.
But we've got args on **both** CLI and class that should make it easy to only give `_granian` the desired `DEBUG` level, so should be fine.

I'm probably used to seeing this information from Gunicorn baseline
```
[2024-08-02 12:26:58 +0200] [57974] [INFO] Worker reloading: /Users/AmonKhavari/gunicorn/settings.py modified
```
which is why I would've liked to see it on the default log level (-> `INFO`).

In the end it's your call of course :)

gi0baro:
@iamkhav makes sense. Let's merge it as it is and see how it goes.

Issue #34, opened 2023-01-16
Support `max-requests` and `max-requests-jitter` config parameters


sidmitra:
One of the common problems with Django apps is memory leaks. The most common suggestion is to allow the webserver to cycle processes after serving some fixed number of requests. This bypasses memory issues by just returning used up memory and start fresh. 

Perhaps you can consider supporting something similar to gunicorn early in the design.

```bash
gunicorn --max-requests 1000 --max-requests-jitter 50 app.wsgi
```

This is also supported in uwsgi, one of the popular alternatives to gunicorn.
Similar issues are solved in celery `worker_max_tasks_per_child` since i think it's fairly hard to track down and fix memory issues in larger Python projects with hundreds of dependencies. 

References
------------
- https://docs.gunicorn.org/en/latest/settings.html#max-requests
- https://uwsgi-docs.readthedocs.io/en/latest/Options.html?highlight=ma-requests#max-requests
- https://adamj.eu/tech/2019/09/19/working-around-memory-leaks-in-your-django-app/

gi0baro:
This will require quite an effort, as the main process – which right now is quite simple – should behave like a real arbiter and effectively manage child process continuously.

Probably I can take a look at what gunicorn does, but this for sure will require some time.

snuderl:
Similar/related thing is having a request timeout after which to kill a process.

https://docs.gunicorn.org/en/stable/settings.html#timeout

gi0baro:
I'm not sure Granian should handle all of this complexities.
If the aim would be having 1:1 feature comparison with Gunicorn, probably it would make more sense to implement Granian workers for Gunicorn..

sidmitra:
>probably it would make more sense to implement Granian workers for Gunicorn..
This also makes sense, but i'm not aware of the tradeoffs. 

I created the ticket thinking of the feature that has made me choose gunicorn over others like Daphne even. In any reasonably complex web app(think enterprise!) memory leaks seem to be a big issue and hence this seemed like an must-have. This is obviously not true for every project, and perhaps Granian works perfectly fine for other use cases.
It might be possible to keep the design open in a way that these can be implemented by the community perhaps in the future if needed.

cirospaciari:
> I'm not sure Granian should handle all of this complexities. If the aim would be having 1:1 feature comparison with Gunicorn, probably it would make more sense to implement Granian workers for Gunicorn..

timeouts, and backpressure (max req) sound like features for Emmett, not for Granian.

[...]

PR #372, opened 2024-08-04
Add workers lifetime option

gi0baro:


Zerotask:
That's a really nice addition. If I may suggest, could you also add another option for `The maximum amount of requests a worker will be kept alive before respawn` (#34)

mikeckennedy:
Hey @gi0baro this is great. I've been waiting for this feature. 

One enhancement that I think would be really good, would be to add a jitter across workers.

Suppose you have 8 workers and set the restart time to 86,400 (1 day). After one day, as far as I can see, it will stop and restart all 8 workers at roughly the same time. I would much rather have a little bit of time between them to keep the jarring effect of a complete restart happening all at once. 

Could we get something like a `--workers-lifetime-jitter` which would allow us to specify an interval to vary the restart across.

For example: `--workers-lifetime-jitter 60` would have the workers restart across 60 seconds rather than all at once or just each one restart based on random(0, 60) seconds from when it started. 

Anything like this would be good I think (unless there is a default I didn't see in there somewhere).

gi0baro:
Hey @mikeckennedy 👋 

The `--workers-lifetime` uses the *respawn strategy*, so it will respawn all the expired workers sequentially (starting the new one) respecting the `--respawn-interval` option (then shutting down the old one).
So if you just want to wait for a certain amount of time between the workers respawn, you can tune the `--respawn-interval` option. Eg: setting `--workers-lifetime 86400 --respawn-interval 30` will effectively make Granian wait 30 seconds before shutting down the old worker and thus proceed with the respawn of the next one.

Does this work for your use-case?

mikeckennedy:
@gi0baro Perfect, that's more than enough for me. Thanks!

PR #426, opened 2024-11-03
On reload, ignore non python files by default

artefactop:
Testing granian I realized that reload was trigger by default for all kind of files. This is a bit annoying for development if you are just changing the html templates and I think it only needs to reload for Python files.

To do not break the previous behaviour (completely), now the reload only use `DefaultFilter` if params required to build that filter are set.

Happy to add some tests but I couldn't find any similar or related test to guide me in the process.

BTW, from my point of view, reload only on Python files always should be the enough, I don't know if the usage of `DefaultFilter` is a bug or a desired behaviour.

One more think, I couldn't do this without #119 it's my first time working with maturing :)

### DIFF
--- a/granian/server.py
+++ b/granian/server.py
@@ -643,7 +643,7 @@ class Granian:
             reload_filter = watchfiles.filters.DefaultFilter(
                 ignore_dirs=ignore_dirs,
                 ignore_entity_patterns=ignore_entity_patterns,
-                ignore_paths=ignore_paths
+                ignore_paths=ignore_paths,
             )
         else:
             # Use given or python filter rules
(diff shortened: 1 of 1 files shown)


monosans:
This will break reloading for Django developers, since there Granian also has to reload on `.html` files. Please see my suggestion - #437. I think my solution is more flexible.

gi0baro:
Thank you for your contribution @artefactop.

I'd rather avoid making general assumptions based on a single application framework – Django in this case, but can be anything – and just state **this is a breaking change**, as you can't really tell in advance wether the default filtering for watchfiles is actually used by any RSGI, ASGI, WSGI application using Granian out there, which is not necessarily only about Python files (both of you and @monosans mentioned html files, but what about static files like js, css, images?).

So I'm quite on the idea to reject this, and probably be in favour of something act to simplify the current filtering customisation (both in Python and from the CLI). #437 might be a good discussion point for that.

gi0baro:
> When you are developing with any Python framework you don't need to restart your server on static file changes because they are STATIC and **will be served by your server**. I could not figure out any case when you need to restart your server due to changes in static files. 

Can you please lower down your tone a bit? I don't find it productive in any sense.
I can think of at least one case where you have a separate process transpiling javascript files into a bundled/minified version during development and you want to restart the server because of E-tag/caching.

And again, this is not just about Django.

> I see the point that this is a breaking change, feel free to close this PR.

Done. Feel free to open a different PR/discuss about a different implementation in #437.

artefactop:
Sorry, I didn't want to be rude.

PR #438, opened 2024-11-21
fix: KeyError if `_granian` logger is missing in user logging configuration

monosans:
Minimal example:

```python
from granian import Granian

Granian(target="", log_dictconfig={"loggers": {}}).serve()
```


gi0baro:
@monosans thank you for your contribution! ❤️

PR #439, opened 2024-11-21
feat: add support for logging.NOTSET level

monosans:
From Python docs: 

logging.NOTSET: When set on a logger, indicates that ancestor loggers are to be consulted to determine the effective level. If that still resolves to NOTSET, then all events are logged.

### DIFF
--- a/granian/log.py
+++ b/granian/log.py
@@ -14,6 +14,7 @@ class LogLevels(str, Enum):
     warn = 'warn'
     info = 'info'
     debug = 'debug'
+    notset = 'notset'
 
 
 log_levels_map = {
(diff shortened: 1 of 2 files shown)


granian/log.py
```diff
@@ -23,6 +24,7 @@ class LogLevels(str, Enum):
     LogLevels.warn: logging.WARN,
     LogLevels.info: logging.INFO,
     LogLevels.debug: logging.DEBUG,
+    LogLevels.notset: logging.NOTSET,
 }
```
monosans, line 28: We can remove `log_levels_map` at all and use the LogLevels enum above as a mapping. Let me know if you are interested.

gi0baro:
@monosans thank you for your contribution ❤️ 

Gonna merge this as soon as possible, with the 1.7 schedule.

PR #461, opened 2024-12-09
add restart tests

hendrikmuhs:
Spin-off from #435

This change adds restart tests for SIGHUP handling and graceful worker restart.

**Note**: In #435 we agreed on not adding these tests, this PR just leaves them here as example how (graceful) restarts can be tested


gi0baro:
Close as stale

PR #477, opened 2024-12-28
Fixed incorrect parsing of ipv6 addresses in RSGI/WSGI access logs

jyelloz:
Granian is setting REMOTE_ADDR values using SocketAddr::to_string() which uses the format `<host>:<port>` and ipv6 addresses are delimited internally by colons, so the port that needs to be stripped must be taken from the right and there must only be one split done to avoid splitting the ipv6 address.

A possibly simpler alternative is to update Granian's rust wsgi adapter to supply only the client IP address in REMOTE_ADDR so that no parsing needs to be done in the python code.

### DIFF
--- a/granian/rsgi.py
+++ b/granian/rsgi.py
@@ -107,7 +107,7 @@ def _build_access_logger(fmt):
         logger(
             t,
             {
-                'addr_remote': scope.client.split(':')[0],
+                'addr_remote': scope.client.rsplit(':', 1)[0],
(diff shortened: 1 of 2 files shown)


gi0baro:
@jyelloz thank you for your contribution!
Can you do the same also for RSGI (https://github.com/emmett-framework/granian/blob/master/granian/rsgi.py#L110)?

jyelloz:
Sure thing, I just updated the RSGI module with the same change.

Issue #490, opened 2025-01-24
Granian crashes with a UTF-8 error when you send it %c0


ritesh:
Issue: Granian crashes with a UTF-8 error when you send %c0 to an http endpoint.

To reproduce:

I used uv to install and set up a new project

```
$ mkdir /tmp/foo && cd /tmp/foo
$ uv init
$ uv add granian
```

This is my pyproject.toml
```
[project]
name = "foo"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.10"
dependencies = [
    "granian>=1.7.5",
]
```

Contents of main.py
```
async def app(scope, receive, send):
    assert scope['type'] == 'http'

    await send({
        'type': 'http.response.start',
        'status': 200,
        'headers': [
            [b'content-type', b'text/plain'],
        ],
    })
    await send({
        'type': 'http.response.body',
        'body': b'Hello, world!',
    })
```

Run the server in one terminal window
```
$ RUST_BACKTRACE=1 uv run granian --interface asgi main:app
[INFO] Starting granian (main PID: 3951)
[INFO] Listening at: http://127.0.0.1:8000
[INFO] Spawning worker-1 with pid: 3953
[WARNING] ASGI Lifespan errored, continuing without Lifespan support (to avoid Lifespan completely use "asginl" interface)
[INFO] Started worker-1
[INFO] Started worker-1 runtime-1
```

In another terminal window
```
$ curl localhost:8000/%c0
```

This causes the server to crash with an unrecoverable error

```
[INFO] Starting granian (main PID: 3951)
[INFO] Listening at: http://127.0.0.1:8000
[INFO] Spawning worker-1 with pid: 3953
[WARNING] ASGI Lifespan errored, continuing without Lifespan support (to avoid Lifespan completely use "asginl" interface)
[INFO] Started worker-1
[INFO] Started worker-1 runtime-1
thread '<unnamed>' panicked at src/asgi/callbacks.rs:157:9:
called `Result::unwrap()` on an `Err` value: Utf8Error { valid_up_to: 1, error_len: Some(1) }
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
[ERROR] Unexpected exit from worker-1
[INFO] Shutting down granian
```

Expected behaviour: I would expect the server to return an HTTP error code and continue running.

ritesh:
I tried to change the way paths are processed using 
```
[...]

PR #492, opened 2025-01-24
Possible fix for handling invalid http paths

ritesh:
Hello! This is an incomplete attempt at trying to fix the bug that we encountered. I was unable to `cargo build` the project as I'm not familiar with py03. 

What I *think* is happening is that when the path is parsed the `percent_decode` function, it returns a Result type and granian is currently `unwrapping` the result, causing the error to cause the the process to crash. 

If instead of unwrapping it, we return `unwrap_or` with a `Cow<_, str>` with an empty value that should fix it (which is what I've done). Although I suspect this may not be the best way to approach it as there may be other instances of path parsing where we need to decode valid UTF8 but return either an empty string or some HTTP client error? 

Issue: https://github.com/emmett-framework/granian/issues/490

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -117,13 +117,3 @@ asyncio_mode = 'auto'
 
 [tool.uv]
 package = false
-
-[dependency-groups]
-dev = [
-    "httpx>=0.25.2",
-    "maturin>=1.8.1",
-    "pip>=25.0",
-    "pytest>=7.4.4",
-    "pytest-asyncio>=0.21.2",
-    "websockets>=11.0.3",
-]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -117,3 +117,13 @@ asyncio_mode = 'auto'
 
 [tool.uv]
 package = false
+
+[dependency-groups]
+dev = [
+    "httpx>=0.25.2",
+    "maturin>=1.8.1",
+    "pip>=25.0",
+    "pytest>=7.4.4",
+    "pytest-asyncio>=0.21.2",
+    "websockets>=11.0.3",
+]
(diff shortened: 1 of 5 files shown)


ritesh:
That makes sense. 100% agree that users should get a 400 error instead of the path being mutated/mangled. 

Have you also considered if trying to decode it lossily might work? The rationale here would be that the server tries to decode it and assumes it's a user error in encoding the URL. If garbage comes out, you get a 404 since that _invalid path_ doesn't exist. Again, I don't know what the spec says you should do in these cases (i.e. complain about invalid UTF-8 or just say the path doesn't exist). 

If you get garbage UTF-8 sequences, you the infamous UTF-8 replacement character. 

https://docs.rs/percent-encoding-rfc3986/latest/percent_encoding_rfc3986/struct.PercentDecode.html#method.decode_utf8_lossy

gi0baro:
> > Have you also considered if trying to decode it lossily might work? The rationale here would be that the server tries to decode it and assumes it's a user error in encoding the URL. If garbage comes out, you get a 404 since that _invalid path_ doesn't exist. Again, I don't know what the spec says you should do in these cases (i.e. complain about invalid UTF-8 or just say the path doesn't exist).
> 
> I think the best option here is to check ASGI spec and/or what other servers do in order to drive the final decision.

Seems like bot uvicorn and hypercorn do that. Thus, I'm fine switching to utf8 lossy here.

ritesh:
Sounds good - thanks for the info regarding setting up a dev environment.  It is a trivial change but I'd like to add a test or two and learn more about this whole rust in python business. Hope to get it done today if I get some focused time away from work, but if not it will likely be early next week. If anyone else beats me to the punch that's OK too.

ritesh:
I've made an attempt at fixing the failing tests but I'm not sure if this is the right approach.
The server returns a 500 instead of a 404, and I'm not sure where to make the change where invalid UTF-8 paths are treated as 404s.

gi0baro:
@ritesh hey, sorry I directly intervened on the code here, but I wanted this to be merged for the next patch release

Thank you for your contribution

PR #579, opened 2025-05-15
Review shared socket implementation

### LINKED ISSUE #567: Enhancing load handling: solving socket backlog issues during worker reloads under extreme pressure
The change introduced in #544 has significantly improved performance under load, however a nasty side effect was discovered during constant high load – when socket backlogs are non-empty and workers are reloading. This kind of load is quite not typical, but it happens sometimes during surges or botnet attacks – that's why I didn't notice it earlier :(

The problem is that every worker has an _independent_ backlog due to dedicated sockets; i.e., once a worker is reloading (or is killed/exits for any reason), its backlog is simply dropped, and all pending requests are rejected with ECONNRESET.

The (Linux) kernel is supposed to handle this gracefully when SO_REUSEPORT is set, but unfortunately, that is not the case (not sure about other systems).

The solution that mostly mitigates the effect is to revert to a shared listener socket (with `bind()` performed only in the managing process) _but_ still to issue `listen()` independently in every worker, to keep the even load sharing.

This way, the backlog is also _shared_ between workers, and if one is exiting, it does not drop any pending connections – this almost entirely solves the issue, with only a few exceptions (probably system dependent).

The managing process _may_ also issue `listen()` to "buffer" accumulating connections (as it doesn't do `accept()`) if there is _only one_ worker and it is reloading, or even if during chain reload some workers could not keep up. It won't save the world if the backlog gets full; however, for less loaded cases, it will help, allowing for truly graceful and seamless reloads without losing any established connections.

gi0baro:
Close #567


aldem:
The core issue seems to be fixed (no lost or refused connections), however, something is off with process management - when I start many processes (8-16), and then just try to stop the app (no requests served), sometimes some workers just hanging around without exiting. This is reproducible, but unpredictable - may take few runs before it happens.

When I run my load test, and workers are killing themselves after N requests, there is a message like `[ERROR] Unexpected exit from worker-16` (which is normal and expected) - but the worker itself (the process) is still alive and not accepting requests, just hanging, while the app inside is not receiving lifetime event. This way, eventually all of them are effectively dead but hanging around.

And sometimes, when workers hang, they could not be killed with SIGTERM - only hard SIGKILL helps. They don't use CPU though, so this is not busy loop.

gi0baro:
@aldem I'm confused, as this PR doesn't really change processing management. do you see a different behaviour with master?

aldem:
@gi0baro Now, I am confused... It turns out that the problem occurs only when I run it in the standard (OS default) Python's environment, but it doesn't occur in my usual venv (which wasn't active yesterday when I did the test). However,  this is definitely another issue and outside the current scope (I will investigate it later).

Otherwise, it works as expected - no lost connections, no broken requests while workers are reloaded - I believe it is OK to merge.

Thank you!

PR #592, opened 2025-06-02
Revise ASGI ws accept flow to wait for send

### LINKED ISSUE #591: Reproducible ASGI Flow Error when websocket is accepted from background task
Hi,

I've created a minimal reproduction of this issue here:

https://github.com/jthorniley/granian-ex

This is an ASGI application that echos the websocket input.

```
import asyncio


async def dispatch(incoming_message, send):
    if incoming_message["type"] == "websocket.connect":
        print("WebSocket connected")
        await send({"type": "websocket.accept"})
    elif incoming_message["type"] == "websocket.receive":
        data = incoming_message.get("text", "")
        print(f"Received: {data}")
        await send({"type": "websocket.send", "text": f"Echo: {data}"})
    else:
        print(f"Unknown message type: {incoming_message['type']}")


async def app(scope, receive, send):
    if scope["type"] == "websocket":

        tasks = []
        try:
            while True:
                incoming_message = await receive()
                if incoming_message["type"] == "websocket.disconnect":
                    return
                tasks.append(asyncio.create_task(dispatch(incoming_message, send)))

        finally:
            await asyncio.gather(*tasks)
            print("All tasks completed")
```

The issue is that the websocket handler code is run as an asyncio Task, which means that the main loop continues to the next `receive` call (potentially) before the first `send` (responding to the initial `websocket.connect` message) has run.

When I run the server code with granian, and use `client.py` to trigger it, `WebSocket connected` is printed multiple times, then the echo code runs, then an exception is raised:

```
> uv run granian --interface=asginl asgi:app
[INFO] Starting granian (main PID: 41481)
[INFO] Listening at: http://127.0.0.1:8000
[INFO] Spawning worker-1 with PID: 41483
[INFO] Started worker-1
[INFO] Started worker-1 runtime-1
WebSocket connected
WebSocket connected
WebSocket connected
Received: Hello world!
[ERROR] Application callable raised an exception
Traceback (most recent call last):
  File "/Users/james/granian-ex/.venv/lib/python3.13/site-packages/granian/_futures.py", line 15, in future_watcher
    await inner(watcher.scope, watcher.proto)
  File "/Users/james/granian-ex/asgi.py", line 29, in app
    await asyncio.gather(*tasks)
  File "/Users/james/granian-ex/asgi.py", line 8, in dispatch
    await send({"type": "websocket.accept"})
RuntimeError: ASGI flow error
```

If I use `uvicorn` or `daphne` there is no such duplication of of the connected message, and there are no errors:

```
> uv run daphne asgi:app
2025-06-02 13:16:16,245 INFO     Starting server at tcp:port=8000:interface=127.0.0.1
2025-06-02 13:16:16,245 INFO     HTTP/2 support not enabled (install the http2 and tls Twisted extras)
2025-06-02 13:16:16,245 INFO     Configuring endpoint tcp:port=8000:interface=127.0.0.1
2025-06-02 13:16:16,246 INFO     Listening on TCP address 127.0.0.1:8000
127.0.0.1:50366 - - [02/Jun/2025:13:16:17] "WSCONNECTING /" - -
WebSocket connected
127.0.0.1:50366 - - [02/Jun/2025:13:16:17] "WSCONNECT /" - -
Received: Hello world!
127.0.0.1:50366 - - [02/Jun/2025:13:16:17] "WSDISCONNECT /" - -
All tasks completed

> uv run uvicorn asgi:app
INFO:     Started server process [53188]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
WebSocket connected
INFO:     127.0.0.1:50757 - "WebSocket /" [accepted]
INFO:     connection open
Received: Hello world!
All tasks completed
INFO:     connection closed
```

I think in https://github.com/emmett-framework/granian/blob/02f683950c22f962a7cd9c04851c8dd6a3cd1c30/src/asgi/io.rs#L430 - a websocket.connect message is returned immediately if the connection has not yet been accepted, but this doesn't account for the app server potentially calling `receive` again before `send`ing the `websocket.accept` message.

This pattern is how Django Channels works so I suspect is the cause of issues relating to ASGI Flow Error with Channels e.g. https://github.com/emmett-framework/granian/issues/186 but can't be certain about other cases.

gi0baro:
Close #591

### DIFF
--- a/src/asgi/io.rs
+++ b/src/asgi/io.rs
@@ -46,7 +46,7 @@ pub(crate) struct ASGIHTTPProtocol {
     body_tx: Mutex<Option<mpsc::UnboundedSender<body::Bytes>>>,
     flow_rx_exhausted: Arc<atomic::AtomicBool>,
     flow_rx_closed: Arc<atomic::AtomicBool>,
-    flow_tx_waiter: Arc<tokio::sync::Notify>,
+    flow_tx_waiter: Arc<Notify>,
     sent_response_code: Arc<atomic::AtomicU16>,
 }
 
@@ -307,7 +307,9 @@ pub(crate) struct ASGIWebsocketProtocol {
     upgrade: Mutex<Option<UpgradeData>>,
     ws_rx: Arc<AsyncMutex<Option<WSRxStream>>>,
     ws_tx: Arc<AsyncMutex<Option<WSTxStream>>>,
-    accepted: Arc<atomic::AtomicBool>,
+    init_rx: atomic::AtomicBool,
+    init_tx: Arc<atomic::AtomicBool>,
+    init_event: Arc<Notify>,
     closed: Arc<atomic::AtomicBool>,
 }
 
@@ -325,7 +327,9 @@ impl ASGIWebsocketProtocol {
             upgrade: Mutex::new(Some(upgrade)),
             ws_rx: Arc::new(AsyncMutex::new(None)),
             ws_tx: Arc::new(AsyncMutex::new(None)),
-            accepted: Arc::new(false.into()),
+            init_rx: false.into(),
+            init_tx: Arc::new(false.into()),
+            init_event: Arc::new(Notify::new()),
             closed: Arc::new(false.into()),
         }
     }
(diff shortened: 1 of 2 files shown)


shamoon:
Awesome thank you!

PR #631, opened 2025-07-10
Fix the serve reload filter

### LINKED ISSUE #630: Custom reload_filter may lose functionality
https://github.com/emmett-framework/granian/blob/3f0bbd271d2d2b07405f82d5493ddd1805e184e6/granian/server/common.py#L460C9-L460C31

The `reload_filter` is always updated to `DefaultFilter`.

https://github.com/samuelcolvin/watchfiles/blob/e3c956b2a186433c6c3bb7d27b3d57fe639f9f4d/watchfiles/filters.py#L124

The PythonFilter `__call__` method is also overridden, which will not work in granian.

wu-clan:
Fix #630.

### DIFF
--- a/granian/server/common.py
+++ b/granian/server/common.py
@@ -449,18 +449,14 @@ class AbstractServer(Generic[WT]):
             sys.exit(1)
 
         # Use given or default filter rules
-        reload_filter = self.reload_filter or watchfiles.filters.DefaultFilter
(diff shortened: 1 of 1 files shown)


gi0baro:
This is incorrect, as `reload_filter` should be a class, not an instance. See #634 and also [my comment in the original issue](https://github.com/emmett-framework/granian/issues/630#issuecomment-3072772915)

PR #640, opened 2025-07-21
Add workers max RSS

### LINKED ISSUE #638: Add `--workers-max-rss` to restart workers on memory usage limit
As discussed in https://github.com/emmett-framework/granian/issues/34 , having a `--workers-max-rss` setting would help mitigate possible memory leaks.

I imagine this should check the worker memory usage and after it exceeds this limit, sends a clean shutdown (so it finishes handling whatever requests).

gi0baro:
Close #638


zffocussss:
@gi0baro 
I am a user of gunicorn + uvicorn + Fastapi.I want to replace them with your granian.When will the next release come, including this pr?

PR #646, opened 2025-07-24
Precompressed static files 

### LINKED ISSUE #577: Feature request: Pre-compressed static files
Thanks so much for the static files feature; it simplifies deployment for my hobby project!

I think it would be nice to have support for pre-compressed static files in the static file server, like nginx's [`gzip_static`](https://nginx.org/en/docs/http/ngx_http_gzip_static_module.html#gzip_static) directive.

When enabled, a request matching a static file with an appropriate `Accept-Encoding` header would also check for files matching the path with a `.gz`, `.br`, or `.zst` suffix and respond with the smallest matching compressed version and appropriate `Content-Encoding` and `Vary` headers.

Please let me know if you think this feature makes sense for Granian. I'm still new to Rust, but I'd be happy to make a PR if interested!

mollymorphous:
This adds support for precompressed sidecars to the static file server. Use `--static-path-precompressed` to enable.

When enabled, the static file server checks for the presence of sidecar files with `.br` and `.gz` extensions, and serves them to clients which advertise support for Brotli or gzip Content-Encoding. The base uncompressed file must also exist.

This implementation tries reading the sidecars sequentially followed by the uncompressed base file. This adds a minor overhead for intrinsically compressed files without sidecars such as images and media. A future optimization could include a list of common file extensions to exclude. 

Closes #577


gi0baro:
> What do you think about fallback encodings? I think there's two main options:
> 
>     1. Always check brotli, then gzip, then none. This has the disadvantage that files without sidecars cost three lookups for brotli-capable clients (all browsers). But the advantage of being able to use a gzip sidecar if a brotli one wasn't present.
> 
>     2. Only check the "best" encoding the client supports, then fallback to none. This saves us a lookup on files without sidecars, but it does mean that every precompressible file needs a brotli sidecar to work with modern browsers.

I'd say to go with the first option, and we can clarify in the project's README what's the *hidden cost* of enabling pre-compressed assets. 
Then later on we can think of ways to optimize that check, a few that comes to my mind right now:
1. adding an option in granian to specify only specific formats to lookup (eg: `--static-path-precompressed-ext gz` would only do that check, not brotly)
2. pre-scanning the static mount path on start to build a hashmap of the files and avail sidecars, so the matcher already has the answer in memory
3. feeding a hashmap of the files once we check them to already have the result for the next request
4. combine 2 and 3

But I'd say that's worth a separate discussion after this PR gets merged, there's no need to have *the optimal* impl from day 0 :)

gi0baro:
@mollymorphous any updates to share?

mollymorphous:
Hi @gi0bro, I don't have capacity to work on this right now. Thanks for your consideration!

PR #658, opened 2025-08-12
add log atoms

zffocussss:
- http request header
- http request body size

### DIFF
--- a/README.md
+++ b/README.md
@@ -384,15 +384,16 @@ The following atoms are available for use:
 | identifier | description |
 | --- | --- |
 | addr | Client remote address |
-| time | Datetime of the request | 
+| time | Datetime of the request |
 | dt_ms | Request duration in ms |
 | status | HTTP response status |
 | path | Request path (without query string) |
 | query\_string | Request query string |
 | method | Request HTTP method |
+| {header}i| Request HTTP header|
 | scheme | Request scheme |
 | protocol | HTTP protocol version |
-
+| rbl | HTTP request body length (wsgi only)|
 ### Workers and threads
 
 Granian offers different options to configure the number of workers and threads to be run, in particular:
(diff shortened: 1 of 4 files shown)


zffocussss:
@gi0baro can you help have a check?

gi0baro (changes requested):
@zffocussss thank you for your contribution.

Unfortunately, I can't merge this due to the following reasons:
- it's missing the RSGI protocol implementation. Any change in Granian regarding shared interfaces should be performed on all protocols
- I find the WSGI implementation to be very *hacky* (eg: manually patching the scope to store data), and I'm confused on why you `del` stuff from objects
- The ASGI implementation is incomplete
- The *response size* wouldn't be accurate 
- The overall implementation might have severe performance implications (parsing all headers just to print one or few, usage of `re.match`, `SafeAtoms` wrapper, double iteration of the response, etc)

So in general, I think this should be designed quite differently.

zffocussss:
> @zffocussss thank you for your contribution.
> 
> Unfortunately, I can't merge this due to the following reasons:
> 
> * it's missing the RSGI protocol implementation. Any change in Granian regarding shared interfaces should be performed on all protocols
> * I find the WSGI implementation to be very _hacky_ (eg: manually patching the scope to store data), and I'm confused on why you `del` stuff from objects
> * The ASGI implementation is incomplete
> * The _response size_ wouldn't be accurate
> * The overall implementation might have severe performance implications (parsing all headers just to print one or few, usage of `re.match`, `SafeAtoms` wrapper, double iteration of the response, etc)
> 
> So in general, I think this should be designed quite differently.

I will consider your comments.I am not a user of rsgi at all.My goal is to make granian work more flexibly than gunicorn + uvicorn any way.
The response size is calculated differently indeed between asgi and wsgi.
Regarding to the performance issues,I agree with you,trying to find a better way to make it.

gi0baro:
Close as stale

Issue #358, opened 2024-07-29
User Documentation with Github


sreenivasulureddysura:

gi0baro:
@sreenivasulureddysura can you clarify this?

iamkhav:
@sreenivasulureddysura, do you mean "programmatic" as in calling Granian from Python instead of cli?

If so, that's true. This way of using Granian is not documented. One would have to check code to see how https://github.com/emmett-framework/granian/blob/0a85e88ca9b18fa37cfe265bd08f0c5e9c25230f/granian/cli.py#L257-L304

I could add this use-case to README.md. 

BTW, are there any plans to have a dedicated documentation website?

gi0baro:
> @sreenivasulureddysura, do you mean "programmatic" as in calling Granian from Python instead of cli?

Correct, I changed the title and the naming comes from Uvicorn (https://www.uvicorn.org/#running-programmatically).

> I could add this use-case to README.md.
> 
> BTW, are there any plans to have a dedicated documentation website?

IMHO at the moment there's still few content for launching a website; probably relying on markdown files within the GH repo is fine at the moment, in the same way we have those for other projects in the organization like [renoir](https://github.com/emmett-framework/renoir/tree/master/docs).

sreenivasulureddysura:
@gi0baro Let's bring all the markdown files and deploy it using the github page. Mkdocs library will help in the deploy it.  If you are okay with Mkdocs I will submit the PR with the basic structure.

gi0baro:
> @gi0baro Let's bring all the markdown files and deploy it using the github page. Mkdocs library will help in the deploy it. If you are okay with Mkdocs I will submit the PR with the basic structure.

Again, there's no enough content/reasons to publish (yet?) a website about Granian. And if and when that become a need, it would probably be a subdomain of emmett.sh website or even if on its own domain, [built in a similar way](https://github.com/emmett-framework/website).

None of the projects in Emmett organization use Mkdocs; I would avoid having a single project with a different toolset/workflow, and if there's any valid reason to change that organization-wide, we should have that discussion separately from this issue.

Let's start by having documentation in place with just markdown files, then post-pone further considerations at the proper time.

sreenivasulureddysura:
@gi0baro `f emmett.sh website or even if on its own domain, [built in a similar way](https://github.com/emmett-framework/website).` this looks good.  Can we have a sub-domain of the emmett.sh for granian that helps? You have lot of content for Granian Just the document to arrange it .

[...]

PR #664, opened 2025-08-21
Document trusted_hosts for wrap_[aw]sgi_with_proxy_headers

hynek:
It took me a second to figure out how it work, so I thought I write it down for everybody. ;)

P.S. You're wayyy past the point where Granian would deserve proper docs. 🤓

### DIFF
--- a/README.md
+++ b/README.md
@@ -384,7 +384,7 @@ The following atoms are available for use:
 | identifier | description |
 | --- | --- |
 | addr | Client remote address |
-| time | Datetime of the request |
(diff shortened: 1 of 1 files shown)


hynek:
pretty sure my docs update didn't break clippy 😇

gi0baro:
@hynek thank you, I'll review this later in the week.

> P.S. You're wayyy past the point where Granian would deserve proper docs. 🤓

#358 is open for contributions 😉

PR #671, opened 2025-08-27
Define __all__ for package init

anuraaga:
I noticed type checkers have trouble with the `granian` package, and it seems to be because of the lack of a `__all__` explicitly defining the exports. We see this is effective as the lint ignores can go away too.

### DIFF
--- a/granian/__init__.py
+++ b/granian/__init__.py
@@ -1,3 +1,5 @@
-from ._granian import __version__  # noqa: F401
-from ._loops import loops  # noqa: F401
-from .server import Server as Granian  # noqa: F401
+__all__ = ['__version__', 'loops', 'Granian']
+
+from ._granian import __version__
+from ._loops import loops
+from .server import Server as Granian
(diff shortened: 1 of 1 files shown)


gi0baro:
Can you be more specific?
Which type checkers?
What troubles?

anuraaga:
I notice it in VSCode, which would be same with pyright

image

```
❯ uv run pyright
Uninstalled 1 package in 355ms
Installed 1 package in 138ms
/Users/anuraag/git/connecpy/example/example/eliza_service.py
  /Users/anuraag/git/connecpy/example/example/eliza_service.py:67:21 - error: "Granian" is not exported from module "granian"
    Import from "granian.server" instead (reportPrivateImportUsage)
1 error, 0 warnings, 0 informations
```

gi0baro (approved):

anuraaga:
@gi0baro PEP style guide suggests dunder variables like this come before imports. Are you sure you want to move like that?

PR #672, opened 2025-08-27
Use monotonic time and perf_counter

kikones34:
When using granian in a production system, we noticed an issue. In the logs, we are printing the `dt_ms` field, and it sometimes reports a clearly incorrect time, such as -15220ms. I took a look at the code and saw you're using `time.time()`. This uses the system clock and is sensitive to spontaneous clock updates outside of our control. Instead, for this use case `time.perf_counter()` should be used, which guarantees a monotonically increasing clock (see https://docs.python.org/3/library/time.html).

I took the liberty of replacing all the usages of `time.time()` with `time.perf_counter()` or `time.monotonic()` where appropriate. Note that in CPython, both `perf_counter` and `monotonic` are actually the same function, but conceptually they are defined differently, so in each case I chose the one I thought better fit at my own discretion.

Take into account that these clocks have an undefined reference point, so they are only meaningful when subtracting their values, which I believe is the only use for all the ones I have changed in this PR.

Hopefully with this change, the reported times will be more accurate.

### DIFF
--- a/benchmarks/ws/benchmark.py
+++ b/benchmarks/ws/benchmark.py
@@ -26,21 +26,21 @@ def _client_redy(target):
 
 
 async def _client_recv(ws, to_recv):
-    t_start = time.time()
+    t_start = time.perf_counter()
     recv = 0
     while recv < to_recv:
         await ws.recv()
         recv += 1
-    return (recv, time.time() - t_start)
+    return (recv, time.perf_counter() - t_start)
 
 
 async def _client_send(ws, messages):
-    t_start = time.time()
+    t_start = time.perf_counter()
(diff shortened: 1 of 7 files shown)


zffocussss:
@kikones34, how does Granian perform in your production system?

kikones34:
@zffocussss it seems to be performing pretty well! Only issue so far is the occasionally inaccurate times reported in the logs.

zffocussss:
> @zffocussss it seems to be performing pretty well! Only issue so far is the occasionally inaccurate times reported in the logs.

I am encouraged by your feedback to apply Granian in my production k8s.

PR #691, opened 2025-09-17
Fix close embed have Rust Fatal error for IO Safety violation 

Noahnut:
# Summary

This PR updates `SocketHolder::as_tcp_listener` and `SocketHolder::as_unix_listener` to use `try_clone()` instead of constructing listeners with `from_raw_fd(as_raw_fd())`.

# Motivation

Using `TcpListener::from_raw_fd(self.socket.as_raw_fd())` (or the equivalent for UnixListener) takes ownership of the same raw file descriptor that self.socket already owns.
This can lead to double-closing the same FD once both objects are dropped, causing runtime aborts such as:

`fatal runtime error: IO Safety violation: owned file descriptor already closed`

# Benefits

Memory/FD safety: each listener now owns a duplicated file descriptor, preventing double-close issues.

Eliminates unsafe: removes the need for unsafe blocks.

Minimal overhead: dup() system call cost is negligible compared to network I/O.

# Notes

This change only affects FD ownership semantics. Runtime behavior of the listeners remains the same, except that it avoids UB/crashes when objects are dropped in different orders.

### DIFF
--- a/src/net.rs
+++ b/src/net.rs
@@ -4,7 +4,7 @@ use pyo3::{IntoPyObjectExt, prelude::*};
 use std::net::{IpAddr, TcpListener};
 #[cfg(unix)]
 use std::os::unix::{
-    io::{AsRawFd, FromRawFd},
+    io::{AsRawFd},
     net::UnixListener,
 };
 #[cfg(windows)]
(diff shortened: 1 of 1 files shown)


gi0baro:
Thanks for your contribution. This is now superseded by #746

PR #696, opened 2025-09-24
Support disabling forced reload

Sube-py:
discussions: https://github.com/emmett-framework/granian/discussions/695
The document(👇🏻) states that the default value is disabled
```md
--workers-kill-timeout DURATION
                                  The amount of time in seconds (or a human-
                                  readable duration) to wait for killing
                                  workers that refused to gracefully stop
                                  [env var: GRANIAN_WORKERS_KILL_TIMEOUT;
                                  default: (disabled); 1<=x<=1800]
```

but the code:
```python
if self.reload_on_changes and self.workers_kill_timeout is None:
      self.workers_kill_timeout = 3.5
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,7 +298,7 @@ Options:
                                   readable duration) to wait for killing
                                   workers that refused to gracefully stop
                                   [env var: GRANIAN_WORKERS_KILL_TIMEOUT;
-                                  default: (disabled); 0<=x<=1800]
+                                  default: (disabled); 1<=x<=1800]
   --factory / --no-factory        Treat target as a factory function, that
                                   should be invoked to build the actual target
                                   [env var: GRANIAN_FACTORY; default:
--- a/granian/cli.py
+++ b/granian/cli.py
@@ -293,7 +293,7 @@ def option(*param_decls: str, cls: Optional[Type[click.Option]] = None, **attrs:
 )
 @option(
     '--workers-kill-timeout',
-    type=Duration(0, 1800),
+    type=Duration(1, 1800),
     help='The amount of time in seconds (or a human-readable duration) to wait for killing workers that refused to gracefully stop',
     show_default='disabled',
 )


Sube-py:
The question is: When I changed the code during debugging, it was forced to reload before the debugging was completed.

First, we should be consistent with the documentation, and second, we should allow users to actively turn off(use 0) this configuration.

gi0baro:
> The question is: When I changed the code during debugging, it was forced to reload before the debugging was completed.

I'm confused, you run Granian with the reloader, but you don't want it to actually reload?

> First, we should be consistent with the documentation, and second, we should allow users to actively turn off(use 0) this configuration.

The code change you proposed would just break the reloader for several people, that's not the solution.

Sube-py:
> I'm confused, you run Granian with the reloader, but you don't want it to actually reload?

Thanks for your follow-up, and I apologize if my previous description was unclear.

To clarify, the change I'm proposing is about the workers-kill-timeout setting — not to disable reload entirely, but to prevent the forced kill behavior when reload times out.

For example, when I’m debugging code with breakpoints, sometimes I might accidentally trigger a file change (e.g., by unintentionally pressing a key), which starts a reload. If the reload times out, the worker process gets forcefully killed — and that interrupts my debugging session.

That’s why I’d like to suggest allowing workers-kill-timeoutto be set to 0, rather than requiring it to be greater than 1. A value of 0 would mean the forced kill step is skipped, giving the process a chance to continue naturally.


> The code change you proposed would just break the reloader for several people, that's not the solution.

I think my changes are in line with the documentation (the default workers-kill-timeout is disabled), rather than giving users a default value (3.5). If the user needs it, they can actively modify it.

Sube-py:
If you have used uvicorn, you will find that even if you change the file when you hit the breakpoint, it will not reload immediately, but will wait until you finish debugging.

Sube-py:
Take a look plz @gi0baro

Sube-py:
> Please revert the changes on `cli.py` and README, your edits to `common.py` are enough to solve the issue you have.

Already revoked 🤝

PR #697, opened 2025-09-25
Set UDS file permission

apapsch:
Refines #97 with configurable file permission for the Unix Domain Socket.


apapsch:
I have reverted the umask change and made file permission optional, as per the comment in #97:

> To my perspective, the initial implementation by @apapsch was correct, I would only make the chmod optional: by default is unset and uses whatever umask the user might have set on the granian process.

ionelmc:
Hello, did this make it into any release yet?

gi0baro:
No, will be part of upcoming 2.6.

PR #717, opened 2025-10-24
Fix ASGI cookie header joining

elohmeier:
Fix the ASGI cookie header handling to match the semantics we just applied for WSGI in PR #683. Rather than emitting one ASGI cookie header entry per incoming header, we now join multiple cookies with ; so downstream frameworks receive the expected value. No other behavior changes.

Tested this successfully with Django using `granian --interface asginl --host 127.0.0.1 --port 8000 --http 2 --ssl-certificate localhost+dev.pem --ssl-keyfile localhost+dev-key.pem my_project.asgi:application`.

### DIFF
--- a/src/asgi/utils.rs
+++ b/src/asgi/utils.rs
@@ -72,7 +72,33 @@ macro_rules! build_scope_common {
         scope_set!($py, $scope, "query_string", PyBytes::new($py, query_string.as_bytes()));
 
         let headers = PyList::empty($py);
+        let cookie_joined = {
+            let mut values = $req.headers.get_all(header::COOKIE).iter();
+            values.next().map(|first| {
+                let mut joined = Vec::new();
+                joined.extend_from_slice(first.as_bytes());
+                for value in values {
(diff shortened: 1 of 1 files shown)


gi0baro:
I'm not going to accept this.
The change you're referring to for WSGI was a bug in Granian do to the fact the protocol requires chaining.
ASGI protocol spec is clear and designed so that applications receive the headers as a list of tuples. Servers shouldn't do chaining or transformation at any point, regardless of the specific header name.
Frameworks failing to process multiple cookie values are not ASGI compliant and should be fixed in that regards.

elohmeier:
Thanks for clarifying. I did not check the ASGI protocol spec, thought it would be similar to WSGI in this regard. I'll try to get a fix into Django then.

elohmeier:
fyi this has been fixed in Django 6.0rc1 - see https://github.com/django/django/blob/97acd4d2f92eef8c285bac070d437bf0fd52e071/docs/releases/6.0.txt#L313

PR #734, opened 2025-11-11
Fix static files with non-ASCII filenames return 404

recloud-dev:


gi0baro:
@recloud-dev thank you for your contribution. I made #744 out of this which implements a simpler set of changes.

PR #738, opened 2025-11-13
Handle workers max RSS and worker max lifetime with reload on

### LINKED ISSUE #737: Running server with --reload shuts down prematurely
When running server with `--reload` does not handle the `--workers-max-rss` and `--workers-lifetime` options properly and the server will die either after `--rss-sample-interval` or `--workers-lifetime` duration because neither the `rss_signal` nor the `lifetime_signal` flag is handled in the `_serve_with_reloader` function.

In case of `--workers-lifetime` a warning message is logged saying it will be ignored with `--reload` enabled, but nothing changes.

simon-the-sloth:
Fixes #737

### DIFF
--- a/granian/server/common.py
+++ b/granian/server/common.py
@@ -439,8 +439,38 @@ class AbstractServer(Generic[WT]):
         self._call_hooks(self.hooks_reload)
         return self._respawn_workers(workers, spawn_target, target_loader, delay=self.respawn_interval)
 
+    def _handle_lifetime_signal(self, spawn_target, target_loader):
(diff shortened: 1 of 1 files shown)


gi0baro:
This is intentionally not supported.
See https://github.com/emmett-framework/granian/commit/39139fdb201c547a55143d2b5baa7ed9b40b27fb for the fix to avoid launching the lifetime checker thread.

simon-the-sloth:
Perfect, thanks for the quick response and fix!

PR #753, opened 2025-11-27
Add missing stub file return type annotations

### LINKED ISSUE #752: `_granian.pyi` stub file missing return types for `None`
If a function returns `None`, it should be explicitly annotated as-such for code to be fully-typed. According to PEP-484 if no return type is provided, `Any` should be assumed. Some type checkers allow enabling leniency for this, but in strict mode do not allow it.

Relevant extracts from PEP 484 ["the meaning of annotations"](https://peps.python.org/pep-0484/#the-meaning-of-annotations):
 
>  For a checked function, the default annotation for arguments and for the return type is `Any`.

> Note that the return type of `__init__` ought to be annotated with `-> None`. The reason for this is subtle. If `__init__` assumed a return annotation of `-> None`, would that mean that an argument-less, un-annotated `__init__` method should still be type-checked? Rather than leaving this ambiguous or introducing an exception to the exception, we simply say that `__init__` ought to have a return annotation; the default behavior is thus the same as for other methods.

oliverlambson:
Closes #752


gi0baro (approved):

gi0baro:

PR #755, opened 2025-11-27
Upgrade type annotations away from 3.9 syntax

### LINKED ISSUE #754: Use of deprecated generic type aliases instead of builtins
Since granain supports >=3.10, use of the generic typing aliases is unnecessary (and they are deprecated, see ["historical and deprecated features"](https://typing.python.org/en/latest/spec/historical.html#generics-in-standard-collections)). e.g. `typing.Tuple` should just be `tuple` from 3.10 onwards.

It's tedious to update this all manually, so I suggest enabling the ruff lint rules for pyupgrade to handle this automatically and check it going forward:

```diff
  # pyproject.toml
  ...
  [tool.ruff]
  line-length = 120
  extend-select = [
      ...
      'W',      # pycodestyle
+     'UP',     # pyupgrade
  ]
```

oliverlambson:
Closes #754


gi0baro:
Thanks for your contribution!

PR #773, opened 2025-12-31
Add SSL/TLS section documenting PKCS#8 requirement

### LINKED ISSUE #760: PKCS#1 support
Hello!

When I try to start the embedded server using PKCS#1 formatted PEM certificates like this:

```python
server = Server(
    app,
    address="0.0.0.0",
    port=8000,
    interface=Interfaces.ASGI,
    ssl_cert=self.cert_file_path(),
    ssl_key=self.key_file_path(),
    ssl_key_password=<password>,
)
```

The server crashes with:
```
thread '<unnamed>' (127691) panicked at src/tls.rs:76:22:
```

I tried the same, but using PKCS#8 format - it works fine.

Is it deliberate decision not to support PKCS#1 format? I do understand PKCS#1 is somewhat legacy, though I did not found any information whether granian does or does not support PKCS#1 by design.

Thank you!

JaeHyuckSa:
Add SSL/TLS section to README noting that private keys must be in PKCS#8 format (PKCS#1 is not supported)

Closes #760

### DIFF
--- a/README.md
+++ b/README.md
@@ -404,6 +404,10 @@ The following atoms are available for use:
 | scheme | Request scheme |
 | protocol | HTTP protocol version |
 
+### SSL/TLS
+
+When configuring SSL/TLS, the private key file must be in PKCS#8 format. PKCS#1 format is not supported.
+
 ### Workers and threads
 
 Granian offers different options to configure the number of workers and threads to be run, in particular:


gi0baro:
Thank you for your contribution, I opted for #780 instead

PR #775, opened 2026-01-01
Fix crash when combining workers_lifetime and reload

blopker:
Hey! I just migrated totem.org to use Granian from Gunicorn. It's wonderful, less memory used and more performance!

I did run into an issue that slowed me down a bit when using `--reload` with `--workers-lifetime` for the dev server though. The order of checks makes it so the server could crash under the right conditions. Just needed to switch the checks!

I tried to add tests, but needed a mock to make it work. Happy to remove them if that's not what you want.

Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 .DS_Store
 *.pyc
 __pycache__
-uv.lock
 
 .idea/
 *.sublime-*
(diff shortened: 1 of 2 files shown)


gi0baro:

blopker:
Thanks! If you're open to suggestions, after poking around the codebase, it's very nice. However, I did notice that this issue would not have happened if more of the code was type annotated.  Although `basedpyright` was able to flag the issue in static analysis without any extra types, but there were a lot of false positives that would require annotations to actually use it on the rest of the code.

But yeah, typing Python is still pretty annoying.

PR #784, opened 2026-01-12
Add option to serve index files for directory listings

tobiasge:
Setting this option will enable Granian to also serve directories from the static file mount using the given file as index file.

### DIFF
--- a/tests/apps/asgi.py
+++ b/tests/apps/asgi.py
@@ -61,7 +61,7 @@ async def echo(scope, receive, send):
 
 
 async def pathsend(scope, receive, send):
-    path = pathlib.Path.cwd() / 'tests' / 'fixtures' / 'media.png'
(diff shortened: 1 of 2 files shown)


tobiasge:
I've a few application where I need to replace Nginx Unit. With this server I could serve a Python application and it's static files with one server.
With this additional option I could use Granian in our Docker images without having to use an additional webserver (and some kind of process supervisor) inside the container. This would make the container setup easier.
If it was implemented with a rewrite logic as in you example above, would I need to set this option for every folder with an index file? Notably in this use case ([the NetBox Docker image](https://github.com/netbox-community/netbox-docker/)) I have the generated documentation from the NetBox project which contains many folders.

gi0baro:
Gotcha. Let me think about this a bit more.

PR #791, opened 2026-01-25
Add static path precompressed option

### LINKED ISSUE #577: Feature request: Pre-compressed static files

oliverlambson:
Closes #577

Carries on the work from #646, changes from that PR:

- adds lazy caching of sidecar file existence
- adds full support for RFC 7231 Accept-Encoding negotiation (client order, `q=`, `*`)
- adds zstd as an encoding type (and prioritises it highest)

---

**example**

```http
GET /static/styles.css HTTP/1.1
Accept-Encoding: br, gzip
```

1. `./static/styles.css.br` - if exists, serve with `Content-Encoding: br`
2. `./static/styles.css.gz` - if exists, serve with `Content-Encoding: gzip`
3. `./static/styles.css` - fallback to uncompressed

```http
HTTP/1.1 200 OK
Content-Type: text/css
Content-Encoding: br
Vary: Accept-Encoding
```

---

**changes to public apis**

```bash
granian app:app \
  --static-path-mount ./static \
  --static-path-route /static \
  --static-path-expires 31536000 \
  --static-path-precompressed  # new (default disabled)
```

```python
from granian import Granian

server = Granian(
    "app:app",
    static_path_mount="./static",
    static_path_route="/static",
    static_path_expires=31536000,
    static_path_precompressed=True,  # new (default false)
)
```


gi0baro:
@oliverlambson thank you for your contribution ❤️ 

While I originally planned to review this for 2.7, given the amount of changes this introduces, and also the fact that 2.7 is already quite a big set of changes, I rescheduled this for 2.8.

Gonna review the code as soon as I can, probably in the following weeks.

oliverlambson:
@gi0baro that sounds great, thanks!

oliverlambson:
@gi0baro thanks for 2.7 release,  have rebased and updated this branch to suppored the new multi-mount config. Chose to keep the compression option as a global since expires and dir-to-file are both global too

gi0baro (commented):
Hey @oliverlambson, here's my first pass on this.

Before going into code-specific things, I would:
- drop the *caching mechanism* entirely. It adds a bunch of code-complexity, and it's not clear at the time being whether its presence is needed or not for performance reasons. I'm up to to reconsider some caching mechanism once the implementation is completed and we actually verify we're not satisfied with the performance we see.
- make the supported encodings an actual parameter configurable by the user. To my perspective it doesn't really makes sense to check for all the possible file formats every time while the user which sets up Granian to serve pre-compressed assets knows which file formats are actually available. With this change, checking for all the formats becomes a *worst case scenario* which the user voluntarily requested. This probably also allows to refactor the amount of nested for-loop checks in the code and use `HashSet` instead.

PR #657, opened 2025-08-11
Expose runtime metrics

butlerx:
Addresses https://github.com/emmett-framework/granian/issues/610

- Add an IPC metric reporter and collector
  - uses the `interprocess` crate to get cross platform support
- Added prometheus endpoint for exposing metrics
- uses the `tokio-metrics` crate to expose tokio metric
- Added the Following metrics
  - `granian_number_workers`
  - `granian_worker_started_total`
  - `granian_worker_stopped_total`
  - `granian_requests_processed_total`
  - `granian_python_call_latency`
  - `granian_http_responses_total`
  - `granian_requests_processed_total`

**Note**: I'm not familiar enough with granian to know how to capture all the metrics listed in the issue, so pointers on where to collect those metrics from would be appreciated

Example of prom endpoint 
```# HELP granian_total_busy_duration The amount of time worker threads were busy
# TYPE granian_total_busy_duration counter
granian_total_busy_duration{worker_id="2"} 0
granian_total_busy_duration{worker_id="1"} 0

# HELP granian_min_busy_duration The minimum amount of time a worker thread was busy
# TYPE granian_min_busy_duration counter
granian_min_busy_duration{worker_id="2"} 0
granian_min_busy_duration{worker_id="1"} 0

# HELP granian_connections_total Total number of connections established by each worker since startup
# TYPE granian_connections_total counter
granian_connections_total{worker_id="2"} 1

# HELP granian_requests_processed_total Total count of HTTP requests processed by each worker since startup
# TYPE granian_requests_processed_total counter
granian_requests_processed_total{method="GET",path="/",worker_id="2"} 3

# HELP granian_http_responses_total Counter of HTTP responses grouped by status code per worker
# TYPE granian_http_responses_total counter
granian_http_responses_total{status_code="2xx",worker_id="2"} 3

# HELP granian_max_busy_duration The maximum amount of time a worker thread was busy
# TYPE granian_max_busy_duration counter
granian_max_busy_duration{worker_id="1"} 0
granian_max_busy_duration{worker_id="2"} 0

# HELP granian_worker_started_total Total number of workers started since the application started
# TYPE granian_worker_started_total counter
granian_worker_started_total 2

# HELP granian_number_workers Current number of active workers in the Granian application
# TYPE granian_number_workers gauge
granian_number_workers 2

# HELP granian_global_queue_depth The number of tasks currently scheduled in the runtime's global queue
# TYPE granian_global_queue_depth gauge
granian_global_queue_depth{worker_id="2"} 0
[...]

PR #792, opened 2026-01-29
Add metrics

### LINKED ISSUE #610: Expose runtime metrics
To improve *observability* and provide better ways to service autoscaling, Granian should collect some runtime metrics and expose them through well known APIs (prometheus/openmetrics at least).

Ideally the following metrics should be available:

- no of requests processed (per worker)
- python call latency (bucket per worker)
- no of active threads in the blocking pool (per worker)
- HTTP response codes count (per worker)?
- blocking threads idle/wait time bucket (per worker)?
- backlog/backpressure/active tasks in Rust runtime (per worker)?

Best option in terms of impl seems to be using the `metrics` crate.

For this to work, some form of IPC is needed between the main process – that will expose those metrics in std formats like prometheus – and the workers – that will actually store the metrics. Thus, the IPC impl should allow the master process to collect and store metrics from all the workers. *Given IPC is generally easier on Unix, we might want to restrict this feature to non-windows envs*.

This was originally requested in #275, and also emerged in discussions like #406

gi0baro:
Close #610 
Supersedes #657

### DIFF
--- a/granian/server/mp.py
+++ b/granian/server/mp.py
@@ -81,8 +81,10 @@ class WorkerProcess(AbstractWorker):
             sock, _sso = sock
             if sys.platform == 'win32':
                 sock = SocketHolder(_sso.fileno())
-            else:
-                _ipc_handle = IPCSenderHandle(ipc.fileno()) if ipc else None
+            elif ipc:
+                _ipc_fd = os.dup(ipc.fileno())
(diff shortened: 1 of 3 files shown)


Spenhouet:
Super nice! Hope this gets into a release soon 😍

PR #802, opened 2026-02-07
Fix potential deadlock in PyFutureAwaitable callback dispatch

### LINKED ISSUE #750: Server freeze with Django Channels and WebSocket under low load
## Environment
- **Granian version:** 2.6.0
- **Python version:** 3.11+
- **Framework:** Django 5.1.5 + Django Channels 4.3.1
- **Channel Layer:** Redis (channels-redis 4.3.0)
- **Deployment:** Railway (cloud platform)
- **OS:** Linux (Railway container)

## Granian Start Command
```bash
granian --interface asgi --host 0.0.0.0 --port $PORT backend.asgi:application
```

## Problem Description
Server becomes completely unresponsive (freezes) after handling a small number of WebSocket messages through Django Channels. Response time degrades to 30+ seconds and **never recovers**, requiring container restart.

## Symptoms
1. ✅ Server starts normally and handles initial WebSocket connections fine
2. ✅ First ~10-15 messages relay successfully (Mobile → Backend → Desktop via WebSocket)
3. ❌ After ~19 messages in a few seconds, response time spikes to 30+ seconds
4. ❌ Server stops responding to **all requests** (HTTP and WebSocket)
5. ❌ **No error logs, no panic messages** - just silent freeze
6. ❌ Container becomes unresponsive until manual restart

## Architecture
Django Channels WebSocket consumers using:
- **Async Redis** connection pool (`redis.asyncio`, max 50 connections)
- **Channel Layer** for message routing between WebSocket consumers
- **Mobile → Backend → Desktop** relay pattern (E2EE message forwarding)

```python
# Simplified consumer example
class RelayConsumer(AsyncWebsocketConsumer):
    async def receive(self, text_data):
        data = json.loads(text_data)
        
        # Forward message to desktop via channel layer
        await self.channel_layer.send(
            desktop_channel,
            {
                'type': 'relay_from_mobile',
                'message': data,
            }
        )
```

## What We Tried (Did NOT Fix)
- ❌ Adjusting `--runtime-threads` (1, 4, 8, 16)
- ❌ Changing `--workers` count
- ❌ Optimizing code (cached channels, removed Redis writes from hot path)
- ❌ Reducing message volume (only 19 messages before freeze)
- ❌ Different Granian flags

## What FIXED It
✅ **Switching to Uvicorn 0.38.0** completely resolved the issue - same code, same load, zero freezes.

```bash
# This works perfectly:
uvicorn backend.asgi:application --host 0.0.0.0 --port $PORT --workers 2 --log-level warning
```

## Logs Before Freeze
```
📨 Message relayed: Mobile → Desktop | User: user@example.com | ⏱️ Total: 15.2ms (cached channel)
📨 Message relayed: Mobile → Desktop | User: user@example.com | ⏱️ Total: 12.8ms (cached channel)
📨 Message relayed: Mobile → Desktop | User: user@example.com | ⏱️ Total: 14.1ms (cached channel)
...
[19 messages total, then complete silence - no more logs]
```

**Railway metrics at freeze time:**
- Memory: around 300-400MB relatively stable
- CPU: Normal
- Network: Normal
- Response time: 50ms → **30,000ms+** (never recovers)

## System Info

<details>
<summary>requirements.txt (relevant packages)</summary>

```
Django==5.1.5
channels==4.3.1
channels-redis==4.3.0
redis==5.2.1
granian==2.6.0
asgiref==3.10.0
uvicorn==0.38.0
websockets==15.0.1
```
</details>

## Reproduction Steps
1. Set up Django Channels with Redis backend
2. Create WebSocket relay consumer (mobile → backend → desktop)
3. Deploy with Granian using ASGI interface
4. Send 15-20 rapid WebSocket messages through relay
5. Observe server freeze after ~19 messages

## Expected Behavior
Server should handle WebSocket messages continuously without freezing (like Uvicorn does).

## Actual Behavior
Server freezes silently after ~19 messages with no error logs, requiring restart.

What I have noticed additionally is that it keeps working relatively fine until i start sending  websocket messages slowly in faster intervals, but it's stil very small  intervals.

Funny that if I run it on may windows machine, this issue doesn't persist or at list I haven't notice.

---

**Additional Context**
We were very excited to use Granian for its performance benefits, but had to switch to Uvicorn for stability. 

Image

ColemanDunn:
fixes #750 

Reproduction of the issue can be found at: https://github.com/ColemanDunn/granian-channels-lock-up

Change is to not hold the RwLock across `call_soon_threadsafe`. The comment is somewhat self documenting of what this change does. I can take it out if needed. Perhaps there is a better fix than this in the underlying code, but I am now unable to reproduce the lockup.

### DIFF
--- a/src/callbacks.rs
+++ b/src/callbacks.rs
@@ -499,23 +499,16 @@ impl PyFutureAwaitable {
             return;
         }
 
-        {
-            // IMPORTANT: do not hold `ack` lock across `call_soon_threadsafe`.
-            // If `call_soon_threadsafe` ever blocks (e.g. write-to-self), holding the lock can
-            // deadlock the event loop thread when it tries to register the callback.
-            let cb_ctx = {
-                let ack = rself.ack.read().unwrap();
-                ack.as_ref()
-                    .map(|(cb, ctx)| (cb.clone_ref(py), ctx.clone_ref(py)))
-            };
-            if let Some((cb, ctx)) = cb_ctx {
-                _ = rself.event_loop.clone_ref(py).call_method(
-                    py,
-                    pyo3::intern!(py, "call_soon_threadsafe"),
-                    (cb, pyself.clone_ref(py)),
-                    Some(ctx.bind(py)),
-                );
-            }
+        if let Some((cb, ctx)) = {
+            let ack = rself.ack.read().unwrap();
+            ack.as_ref().map(|(cb, ctx)| (cb.clone_ref(py), ctx.clone_ref(py)))
+        } {
+            _ = rself.event_loop.clone_ref(py).call_method(
+                py,
+                pyo3::intern!(py, "call_soon_threadsafe"),
+                (cb, pyself.clone_ref(py)),
+                Some(ctx.bind(py)),
+            );
         }
         pyself.drop_ref(py);
     }


gi0baro:
I'm prone to consider this as a _random effect_, other than the actual solution.

I'm not exactly sure how the `RwLock` from `PyFutureAwaitable.ack` can led to deadlocks, given:
- all the calls to code interacting with that lock are performed holding the GIL, thus:
  - none of the involved code can actually access that lock concurrently
  - the main thread is detached from the GIL (with the event loop in idle) when `set_result` gets called
- `call_soon_threadsafe` doesn't really *block*, it simply add the callback to the loop "queue" and write a wake event

I'm going to still review the code and possibly merge it, as it might be a valid concern for the free-threaded builds, but again, it's hard to think this can be the _actual solution_ on whatever the problem in #750 is, and more like a coincidence.

PR #805, opened 2026-02-11
feat: Add support for simultaneous TCP and UDS binding

IngmarStein:
This commit enables Granian to serve on both TCP sockets and Unix Domain Sockets concurrently.

Changes:
- Updated Rust worker implementations (ASGI, RSGI, WSGI) to handle optional dual listeners.
- Enhanced `WorkerConfig` to store both TCP and UDS socket holders.
- Introduced new `serve_*_dual` macros and functions in Rust to manage concurrent listeners using `tokio::join!`.
- Modified Python `cli.py` to allow both `--host` and `--uds` arguments to be specified.
- Updated `AbstractServer` and server implementations (MP, MT) to initialize and pass both sockets to workers.
- Added a new test case `tests/test_dual_socket.py` to verify simultaneous connectivity.
- Updated `granian/_granian.pyi` type stubs to reflect the changes in worker constructors.

### DIFF
--- a/granian/_granian.pyi
+++ b/granian/_granian.pyi
@@ -62,6 +62,8 @@ class ASGIWorker:
         cls,
         worker_id: int,
         sock: Any,
+        uds_sock: Any,
+        ipc: Any,
         threads: int,
         blocking_threads: int,
         py_threads: int,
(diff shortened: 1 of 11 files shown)


gi0baro:
@IngmarStein can you expand on the use-case for this? While I get _what_ this is about, I don't really get the _why_ we would want this.

IngmarStein:
@gi0baro sorry about this, I should have linked to https://github.com/emmett-framework/granian/issues/97#issuecomment-3146659100 as the motivation for the PR!

PR #811, opened 2026-02-13
Implement websocket pings

brain4power:
adapt 'websocket.ping' message_type to Message::Ping


gi0baro:
ASGI protocol doesn't implement a `websocket.ping` message.

Issue #312, opened 2024-06-03
[Feature Request] Access Log User Agent


nitrag:
Requesting that _request_ `UserAgent` and _response_ `BodyLength` (Content size of the response bytes) be added to the access logs.

PR #829, opened 2026-03-26
Access log vars

nitrag:
Refers to https://github.com/emmett-framework/granian/issues/312
                                                                                                                                                
Add support for arbitrary request headers in access log via `%(header{Header-Name})s` atom


nitrag:
@gi0baro ready for review, thanks

yeongzhiwei:
hi @gi0baro , I would also love to have this feature to put it on parity with gunicorn. any progress on this MR?

nalundgaard:
@nitrag does this address the `BodyLength` part of the request? It looks like that is not present in granian.

nitrag:
@nalundgaard I believe there was additional overhead with adding body/content length, so I removed it. I'm hoping this PR is low enough impact so it has higher chance of being merged. 

@gi0baro

gi0baro:
Hey all, please stop pinging me.
If I leave a PR open and don't reject it, it doesn't mean I won't merge it.

A lot happened in my personal life in the past few months – and even if this wasn't the case – give me the time to appropriately review stuff.
When I'll be able to review this, I'll do.

gi0baro (approved):
I refactored this, specifically:
- dropped response headers entirely, too much complexity added for a limited use-case
- use `%(header{Header-Name})s` as the atom, every other atom is not compatible with Gunicorn, thus no reason to use the same fmt just for this one
- removed `user-agent` as a special case, it's an header like everything else
- split all the code paths for header vs non-header cases
- dropped tests, I don't see the value gain on an already slow suite
- updated README accordingly

PR #847, opened 2026-05-06
Fix crash on shutdown

### LINKED ISSUE #846: Crash on shutdown, with metrics enabled
Similar to #799 

- runtime-mode: default
- wsgi, 1 worker, 8 threads
- metrics enabled
- Version: 2.7.2 and 2.7.4
- Env: Linux, base image of python:3.13-slim-trixie

```
[INFO] Shutting down granian
[INFO] Stopping worker-1
"thread '<unnamed>' (7) panicked at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/pyo3-0.27.2/src/internal/state.rs:322:9:
Cannot drop pointer into Python heap without the thread being attached.
```

This crash happening about 10-20% of the time in our Linux env on shutdown.

I built a MRE locally which narrowed down it to metrics enabled/disabled:

Dockerfile
```Dockerfile
FROM python:3.13-slim-trixie
WORKDIR /app
RUN pip install --no-cache-dir granian==2.7.4
COPY main.py .
EXPOSE 8010
ENV RUST_BACKTRACE=full
CMD ["granian", "main:app", \
     "--interface", "wsgi", \
     "--metrics", \
     "--workers", "1", \
     "--blocking-threads", "8", \
     "--host", "0.0.0.0", \
     "--port", "8010"]
```

Running and then Ctrl-C causes crash
```
$ docker build -t gran .
$ docker run -p 8010:8010 "gran"
[INFO] Websockets are not supported on WSGI, ignoring
[INFO] Starting granian (main PID: 1)
[INFO] Listening at: http://0.0.0.0:8010
[INFO] Spawning worker-1 with PID: 8
[INFO] Started worker-1
^C[INFO] Shutting down granian
[INFO] Stopping worker-1

thread '<unnamed>' (7) panicked at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/pyo3-0.27.2/src/internal/state.rs:322:9:
Cannot drop pointer into Python heap without the thread being attached.
stack backtrace:
   0:     0xffff8ceb1cfc - <unknown>
   1:     0xffff8cdeca24 - <unknown>
   2:     0xffff8ceb1558 - <unknown>
   3:     0xffff8ceb1440 - <unknown>
   4:     0xffff8cecf390 - <unknown>
   5:     0xffff8cecf320 - <unknown>
   6:     0xffff8cecf804 - <unknown>
   7:     0xffff8c05245c - <unknown>
   8:     0xffff8c05774c - <unknown>
   9:     0xffff8c203f24 - <unknown>
  10:     0xffff8cbe3ed0 - <unknown>
  11:     0xffff8c4f64a0 - <unknown>
  12:     0xffff8ced0e74 - <unknown>
  13:     0xffff8d575f38 - <unknown>
  14:     0xffff8d5ddc2c - <unknown>
  15:                0x0 - <unknown>
```

askedrelic:
Fixes #846 
Note: generated with Claude assistance

The first commit makes the most sense, to drop the handle. And fixes the crash.

During my testing with profiling, the 2nd commit came up as another race and seems like an optional, larger fix, but makes sense for fixing on most Python versions.

### DIFF
--- a/src/ipc.rs
+++ b/src/ipc.rs
@@ -111,6 +111,7 @@ impl IPCReceiverHandle {
                 Python::attach(|_| {
                     drop(ipc);
                     drop(metrics_aggregator);
+                    drop(handle);
                 });
             });
 


gi0baro:
The first commit is correct.
The 2nd one is just wrong, and it can potentially cause a lot of weird issues.
If you rebase just keeping the first commit I can accept this.

askedrelic:
Okay, removing the 2nd commit.

PR #861, opened 2026-06-15
chore(logging): use concise, professional copy

alessio-locatelli:
Informal language or jokes stop being fun after a day, week, or month of work and turn into clutter. Concise, formal language is the best default.


gi0baro:
[why so serious?](https://youtu.be/PoyejjJGajk?si=JMDTE-Y9OD_5MwKO)

Issue #871, opened 2026-07-02
Add request latency histogram metric


mikmatko:
Currently there's no visibility into request latency.

PR #872, opened 2026-07-02
Add request latency histogram metric

mikmatko:
Solves #871 

Note: AI implemented. I'm not familiar with Rust. I've reviewed the code. I apologize if this is seen as too much AI slop :)

Expose granian_request_duration_seconds, a per-worker Prometheus histogram of application request durations. Static-file requests are excluded.

Adds a LatencyHistogram to WorkerMetrics, a MetricValue::Hist variant for IPC transport, histogram exposition in the aggregator, and integration tests covering exposition, counts, static exclusion and the empty case.

Co-authored-by: GitHub Copilot <copilot@github.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -478,6 +478,7 @@ Granian exposes the following runtime metrics in Prometheus format. All the metr
 | `connections_handled` | counter | absolute number | worker | Number of accepted connections |
 | `connections_err` | gauge | absolute number | worker | Number of failed connections |
 | `requests_handled` | counter | absolute number | worker | Number of processed requests |
+| `request_duration_seconds` | histogram | seconds | worker | Distribution of end-to-end application request durations (until the full response body is sent); static file requests are excluded |
 | `static_requests_handled` | counter | absolute number | worker | Number of processed requests for static files |
 | `static_requests_err` | counter | absolute number | worker | Number of requests for static files resulted in a non 200 response code |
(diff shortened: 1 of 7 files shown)


gi0baro (changes requested):
A few things:
- the implementation is not measuring the actual e2e latency, as on responses wrapping an async stream the time to send the actual bytes is outside that future block
- not sure why the naming for the is `req`/`request` instead of response
- I generally don't allow AI generated code to be merged in Granian, especially if it's added *as-is*, without the proper refinements (eg: the *giant-wall-of-comments* is just useless and awful).

mikmatko:
Thank you for the comments. I've pushed changes.

> * the implementation is not measuring the actual e2e latency, as on responses wrapping an async stream the time to send the actual bytes is outside that future block

I believe streaming responses are now correctly tracked.

> * not sure why the naming for the is `req`/`request` instead of response

OTEL convention, seemingly a "standard" thing for Prometheus metrics.

> * I generally don't allow AI generated code to be merged in Granian, especially if it's added _as-is_, without the proper refinements (eg: the _giant-wall-of-comments_ is just useless and awful).

That's fair :) I've reduced the verbosity of the comments. Personally I think the overly verbose comments are a good thing - it helps a lot when jumping to an unknown codebase. A lot of people read and comprehend natural language faster than code. But yes, it's verbose, maybe even awful, but not useless :)

PR #877, opened 2026-07-07
run pytest and maturin through uv run

### LINKED ISSUE #858: Inconsistent steps in CONTRIBUTING.md
CONTRIBUTING.md instructs that test should be run with:

    $ make build-dev
    $ make test

This doesn't work:

- `build-dev` installs all dependencies in `./.venv/bin/`, which isn't in `$PATH`.
-  `test` runs `pytest` from `$PATH`. This typically won't work. It'll work if you have all the test dependencies installed for the system python/pytest, but will fail otherwise.

`make test` should run `./.venv/bin/pytest` explicitly instead.

EmmanuelNiyonshuti:
Closes #858

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ pysources = granian tests
 build-dev:
 	@rm -f granian/*.so
 	uv sync --group all
-	maturin develop --uv
+	uv run maturin develop --uv
 
 .PHONY: format
 format:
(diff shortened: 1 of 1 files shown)


gi0baro:
Closed in https://github.com/emmett-framework/granian/commit/505130631ee3252a1a40cc94e101a6a0da5f2940

PR #879, opened 2026-07-08
Prevent process environment from overriding WSGI request data

### LINKED ISSUE #878: WSGI: process environment variables override request headers in environ
## What happened

In the WSGI interface, process environment variables override request data in the WSGI `environ`. Any environment variable named `HTTP_<SOMETHING>` (or matching any CGI key like `REQUEST_METHOD`, `CONTENT_TYPE`, ...) silently replaces the corresponding value coming from the actual HTTP request, on every request.

We hit this while migrating a Django app from gunicorn to Granian in Kubernetes: the pod had an environment variable named `HTTP_HOST` (used by the app for unrelated purposes), which replaced the `Host` header of every incoming request. Django's `request.get_host()` then failed with `DisallowedHost` on every request (the env value contained a URL scheme, which is never a valid host), so health probes got HTTP 400 and the deployment crash-looped.

The same mechanism would also let e.g. an `HTTP_AUTHORIZATION` or `HTTP_X_FORWARDED_FOR` environment variable silently replace the request's auth header / client IP.

## Expected behavior

Request data should take precedence over the process environment. This matches the stdlib reference behavior: `wsgiref` uses `os.environ` as the *base* of the environ (`WSGIServer.setup_environ`) and then writes request data *on top of it* (`BaseHandler`). PEP 3333 also describes the CGI variables as coming from the request.

## Reproduction

```python
# app.py
def application(environ, start_response):
    body = environ.get('HTTP_HOST', '<missing>').encode()
    start_response('200 OK', [('content-type', 'text/plain')])
    return [body]
```

```bash
pip install granian==2.7.9
HTTP_HOST=https://evil.example granian --interface wsgi app:application --port 8000 &
curl -H 'Host: localhost:8000' http://127.0.0.1:8000/
# Expected: localhost:8000
# Actual:   https://evil.example
```

## Root cause

`granian/wsgi.py`, `_callback_wrapper`:

```python
basic_env: dict[str, Any] = dict(os.environ)   # os.environ as the base
...
def _runner(proto, scope):
    resp = Response()
    scope.update(basic_env)                    # env applied ON TOP of request scope
```

`scope` arrives from the Rust side already populated with request data (including `HTTP_*` header keys), and `scope.update(basic_env)` then overwrites those keys with the process environment - the opposite of the `wsgiref` precedence.

## Suggested fix

Build the environ with the environment as the base and request data on top, e.g.:

```python
environ = basic_env | scope
```

The `wsgi.*` / `GATEWAY_INTERFACE` / `SERVER_SOFTWARE` / `SCRIPT_NAME` keys set by Granian itself are not present in the request scope, so they are still applied as before; only the request-provided keys stop being overwritten. In a quick microbenchmark this is also at parity with (or slightly faster than) the current `scope.update(basic_env)` since it avoids re-inserting the environment into the request dict.

Barsoomx:
Fixes #878.

## What

In the WSGI interface, `_runner` applied `os.environ` on top of the request scope (`scope.update(basic_env)`), so any process environment variable named like a CGI key (`HTTP_HOST`, `HTTP_AUTHORIZATION`, `REQUEST_METHOD`, ...) silently replaced the corresponding request value on every request.

This PR builds the per-request environ with the environment as the base and request data on top of it (`basic_env | scope`), matching the `wsgiref` precedence (`WSGIServer.setup_environ` + `BaseHandler`).

## Why it matters

Environment variables named `HTTP_*` are not exotic - e.g. a k8s pod with an `HTTP_HOST` env var (a common "public base URL" pattern) breaks the `Host` header of every request, and an `HTTP_AUTHORIZATION` / `HTTP_X_FORWARDED_FOR` env var would silently spoof auth / client IP.

## Notes

- Keys Granian sets itself (`wsgi.*`, `GATEWAY_INTERFACE`, `SERVER_SOFTWARE`, `SCRIPT_NAME`) are absent from the request scope, so they are applied exactly as before.
- The request scope dict is no longer mutated in place; with `url_path_prefix` set, the access log now reports the path as received (prefix included) rather than the stripped one. Happy to preserve the old logging behavior instead if you prefer.
- Performance: microbenchmark of the merge shows `basic_env | scope` at parity with the old `scope.update(basic_env)` for small environments and faster for large ones (it avoids re-inserting the environment into the request dict); an interleaved wrk A/B run (hello-world app, 1 worker) shows no throughput difference beyond noise.
- Test: request header vs. same-named env var - the header must win (fails on master, passes with the fix).

### DIFF
--- a/granian/wsgi.py
+++ b/granian/wsgi.py
@@ -50,7 +50,7 @@ def _callback_wrapper(callback: Callable[..., Any], scope_opts: dict[str, Any], 
         resp = Response()
         environ = basic_env | scope
         if environ['SCRIPT_NAME']:
-            environ['PATH_INFO'] = environ['PATH_INFO'][len(environ['SCRIPT_NAME']) :] or '/'
+            environ['PATH_INFO'] = scope['PATH_INFO'] = environ['PATH_INFO'][len(environ['SCRIPT_NAME']) :] or '/'
 
         rv = callback(environ, resp)
 
--- a/granian/wsgi.py
+++ b/granian/wsgi.py
@@ -48,11 +48,11 @@ def _callback_wrapper(callback: Callable[..., Any], scope_opts: dict[str, Any], 
 
     def _runner(proto, scope):
         resp = Response()
-        scope.update(basic_env)
-        if scope['SCRIPT_NAME']:
-            scope['PATH_INFO'] = scope['PATH_INFO'][len(scope['SCRIPT_NAME']) :] or '/'
+        environ = basic_env | scope
+        if environ['SCRIPT_NAME']:
+            environ['PATH_INFO'] = environ['PATH_INFO'][len(environ['SCRIPT_NAME']) :] or '/'
 
-        rv = callback(scope, resp)
+        rv = callback(environ, resp)
 
         if isinstance(rv, list):
             proto.response_bytes(resp.status, resp.headers, b''.join(rv))
--- a/granian/wsgi.py
+++ b/granian/wsgi.py
@@ -49,8 +49,8 @@ def _callback_wrapper(callback: Callable[..., Any], scope_opts: dict[str, Any], 
     def _runner(proto, scope):
         resp = Response()
         environ = basic_env | scope
-        if environ['SCRIPT_NAME']:
-            environ['PATH_INFO'] = scope['PATH_INFO'] = environ['PATH_INFO'][len(environ['SCRIPT_NAME']) :] or '/'
+        if basic_env['SCRIPT_NAME']:
+            environ['PATH_INFO'] = scope['PATH_INFO'][len(basic_env['SCRIPT_NAME']) :] or '/'
 
         rv = callback(environ, resp)
 
--- a/tests/test_wsgi.py
+++ b/tests/test_wsgi.py
@@ -28,22 +28,6 @@ async def test_scope(wsgi_server, runtime_mode):
 
 
 @pytest.mark.asyncio
-@pytest.mark.skipif(bool(os.getenv('PGO_RUN')), reason='PGO build')
-@pytest.mark.parametrize('runtime_mode', ['mt', 'st'])
-async def test_scope_not_overridden_by_process_env(wsgi_server, runtime_mode, monkeypatch):
-    monkeypatch.setenv('HTTP_X_CUSTOM', 'env_value')
-    monkeypatch.setenv('HTTP_HOST', 'env.example')
-    async with wsgi_server(runtime_mode) as port:
-        res = httpx.get(f'http://localhost:{port}/info', headers=[('x-custom', 'header_value')])
-
-    assert res.status_code == 200
-
-    data = res.json()
-    assert data['headers']['HTTP_X_CUSTOM'] == 'header_value'
-    assert data['headers']['HTTP_HOST'] == f'localhost:{port}'
-
-
-@pytest.mark.asyncio
 @pytest.mark.parametrize('runtime_mode', ['mt', 'st'])
 async def test_body(wsgi_server, runtime_mode):
     async with wsgi_server(runtime_mode) as port:
--- a/tests/test_wsgi.py
+++ b/tests/test_wsgi.py
@@ -28,6 +28,22 @@ async def test_scope(wsgi_server, runtime_mode):
 
 
 @pytest.mark.asyncio
+@pytest.mark.skipif(bool(os.getenv('PGO_RUN')), reason='PGO build')
+@pytest.mark.parametrize('runtime_mode', ['mt', 'st'])
+async def test_scope_not_overridden_by_process_env(wsgi_server, runtime_mode, monkeypatch):
+    monkeypatch.setenv('HTTP_X_CUSTOM', 'env_value')
+    monkeypatch.setenv('HTTP_HOST', 'env.example')
+    async with wsgi_server(runtime_mode) as port:
+        res = httpx.get(f'http://localhost:{port}/info', headers=[('x-custom', 'header_value')])
+
+    assert res.status_code == 200
+
+    data = res.json()
+    assert data['headers']['HTTP_X_CUSTOM'] == 'header_value'
+    assert data['headers']['HTTP_HOST'] == f'localhost:{port}'
+
+
+@pytest.mark.asyncio
 @pytest.mark.parametrize('runtime_mode', ['mt', 'st'])
 async def test_body(wsgi_server, runtime_mode):
     async with wsgi_server(runtime_mode) as port:


gi0baro (changes requested):
A few things:
- I'd rather suggest you to manage your env vars rather than changing 2yo code that thousands of productions deployments out there depend on
- You – should I say Claude? – mention a *microbenchmark* but didn't presented the results. Also quite hard to believe creating a new dictionary per-request has the same performance implications of the current solution (you know, allocations and GC exist)
- If the point of the access log is true, then this contains an additional breaking change, which I'm not willing to push out there
- Based on the description, this should be a 1 line change. It's not, and I fail to find good reasons to have a 24 line diff for this

In general: contributing to Granian kinda requires you to quit the loop for a second and stop delegating your thought process to Claude. If you dislike this, or you're unwilling to do so, the license clearly allows you to maintain your own slopfork.

Barsoomx:
WSL2 VM (kernel 6.6.87.2-microsoft-standard-WSL2) with 6 vCPUs of an AMD Ryzen 7 9800X3D and 9 GB RAM, CPython 3.12.3.

Client and server on the same host over loopback. This is not a quiet bare-metal rig, which is why I report round-to-round variance explicitly below.

microbenchmark (`timeit`) to measure the environ merge in isolation:

- A fresh dict is built per iteration, shaped like the request scope the Rust side hands to `_runner`: 11 request/CGI keys (`SERVER_PROTOCOL`, `SERVER_NAME`, `SERVER_PORT`, `REMOTE_ADDR`, `REQUEST_METHOD`, `PATH_INFO`, `QUERY_STRING`, `wsgi.url_scheme`, `wsgi.input`, `CONTENT_TYPE`, `CONTENT_LENGTH`) plus 15 `HTTP_*` header keys - 26 keys total
- `basic_env` is a prebuilt dict of 30 / 60 / 150 entries (to model small and k8s-sized process environments), constructed once outside the timed loop, exactly like `_callback_wrapper` does at startup
- Each strategy runs 200k iterations under `timeit`. A baseline that only builds the scope dict (no merge) is timed the same way and subtracted, so the figure is the cost of the merge step alone, per request
- Strategies compared: `scope.update(basic_env)` - the code currently on master; the `scope.setdefault(key, value)` loop over `basic_env` (alternative); `basic_env | scope`

The alternative (loop over basic_env) without the access log change:
```diff
     def _runner(proto, scope):
         resp = Response()
-        scope.update(basic_env)
+        for key, value in basic_env.items():
+            scope.setdefault(key, value)
         if scope['SCRIPT_NAME']:
             scope['PATH_INFO'] = scope['PATH_INFO'][len(scope['SCRIPT_NAME']) :] or '/'
```

| basic_env size | `scope.update(basic_env)` (master) | `setdefault` loop           | `basic_env \| scope` |
| -------------- | ---------------------------------- | --------------------------- | -------------------- |
| 30 vars        | 0.36 us                            | 0.86 us                     | 0.42 us              |
| 60 vars        | 0.59 us                            | 1.71 us                     | 0.57 us              |
| 150 vars       | 1.50 us                            | 4.14 us                     | 0.74 us              |

e2e (WSGI section of benchmarks/benchmarks.py)

`oha` 1.14.0 as the load generator (the tool the harness invokes).

`maturin develop --release --features jemalloc`, matching `runbench.sh`.

`granian --interface wsgi --log-level warning --backlog 2048 --no-ws --http 1 --workers 1 --runtime-threads 1 --blocking-threads 1 --runtime-mode mt app.wsgi:app`.

`b1k` (GET returning 1 KB) at concurrency 64, and `echo` (POST 1 KB) at concurrency 64, 4 s primer at c=8, 3 s warm-up, then a measured 10 s run

The variants differ only in `granian/wsgi.py`

5 full rounds for `b1k` and 2 for `echo`.

| `b1k`, c=64, 5 rounds  | mean rps | sd across rounds | per-round values                  |
| ---------------------- | -------- | ---------------- | --------------------------------- |
| master                 | 82.6k    | 4.7%             | 83.4k, 86.7k, 85.7k, 79.8k, 77.6k |
| `setdefault` loop      | 87.1k    | 5.0%             | 85.2k, 92.5k, 90.8k, 82.0k, 85.1k |
| `basic_env \| scope`   | 82.5k    | 3.5%             | 85.9k, 80.5k, 84.3k, 83.2k, 78.7k |

| `echo` 1KB POST, c=64, 2 rounds | mean rps | per-round values |
| ------------------------------- | -------- | ---------------- |
| master                          | 79.8k    | 79.4k, 80.2k     |
| `setdefault` loop               | 81.0k    | 80.2k, 81.7k     |
| `basic_env \| scope`            | 76.5k    | 77.8k, 75.2k     |

No difference outside noise?

The merge step in isolation (timeit): the loop costs +0.5us/request over the C-level `update` with a 30-var environment, +2.6us with 150 vars.

If you want that at zero, `environ = basic_env | scope` benches at parity-to-faster than the current `update` even at 150 vars - building one right-sized dict seems cheaper than re-inserting the whole environment into the request dict on every request - but that's the version with the access-log change.

I'll try re-running the e2e bench with union + additional log fix, since the first run was very noisy

**If you don't want to touch this at all**, I can at least make this a startup warning/error (your decision if this is a warning or an error), when env vars with CGI-style names (`HTTP_*`, `REQUEST_METHOD`, ...) are present in the process env?

so this is at least declared and visible as a contract on startup

Barsoomx:
Re-ran everything on a clean box, since the previous e2e numbers were from a noisy WSL setup.

idle 4 vCPU KVM VPS (CentOS Stream 8, kernel 4.18), CPython 3.12.1, granian @ ca36b7d, `maturin develop --release --features jemalloc`, oha 1.14.0. Nothing else running, load 0.00 before start.

The whole run is scripted, from clone to the result tables: <https://gist.github.com/Barsoomx/677dcb9213aa023e3bb6493eb1f3e6e9>. The script self-checks that the venv actually serves the swapped `wsgi.py`, and I verified each variant behaviorally before benching: with an `HTTP_X_PROBE` env var set and an `x-probe` request header, master returns the env value, all three fixed variants return the header value.

Variants:

- `master` - `scope.update(basic_env)`
- `setdefault` - the loop from the previous comment, in-place, zero per-request allocations
- `union` - `environ = basic_env | scope`, the current PR revision (has the access-log change)
- `union + log fix` - union + a one-line write-back, access log byte-identical to master:

```diff
     def _runner(proto, scope):
         resp = Response()
-        scope.update(basic_env)
-        if scope['SCRIPT_NAME']:
-            scope['PATH_INFO'] = scope['PATH_INFO'][len(scope['SCRIPT_NAME']) :] or '/'
+        environ = basic_env | scope
+        if environ['SCRIPT_NAME']:
+            environ['PATH_INFO'] = scope['PATH_INFO'] = environ['PATH_INFO'][len(environ['SCRIPT_NAME']) :] or '/'

-        rv = callback(scope, resp)
+        rv = callback(environ, resp)
```

Microbench, same method as in the previous comment (timeit, 200k iterations, merge step only, baseline subtracted; this CPU is slower than my previous one, hence bigger absolute numbers):

| basic_env size | `update` (master) | `setdefault` loop | `basic_env \| scope` | `basic_env \| scope` + log fix |
| -------------- | ----------------- | ----------------- | -------------------- | ------------------------------ |
| 30 vars        | 0.94 us           | 1.87 us           | 0.86 us              | 0.88 us                        |
| 60 vars        | 1.68 us           | 3.66 us           | 1.29 us              | 1.16 us                        |
| 150 vars       | 3.26 us           | 8.38 us           | 1.66 us              | 1.69 us                        |

e2e, same harness invocation as the previous comment (`b1k`, c=64), 10 rounds, all variants interleaved within every round. First with the default (small) process environment:

| variant         | mean rps | sd   | vs master | per-round values |
| --------------- | -------- | ---- | --------- | ---------------- |
| master          | 64.2k    | 5.3% | -         | 63.4k, 66.1k, 66.9k, 61.0k, 67.1k, 67.8k, 66.8k, 59.0k, 65.3k, 59.0k |
| `setdefault`    | 63.4k    | 5.7% | -1.4%     | 67.2k, 66.0k, 67.2k, 58.8k, 64.7k, 67.0k, 62.8k, 57.3k, 61.9k, 60.6k |
| `union`         | 64.3k    | 5.8% | +0.0%     | 66.3k, 59.2k, 67.3k, 67.7k, 58.1k, 66.7k, 64.2k, 66.5k, 66.6k, 60.1k |
| union + log fix | 63.7k    | 6.7% | -0.9%     | 67.0k, 58.6k, 68.2k, 65.1k, 58.1k, 67.9k, 57.7k, 67.4k, 61.8k, 65.1k |

All four within +-1.4% with 5-7% sd between rounds of identical code - not resolvable, and the "+5% for setdefault" from my previous WSL run was rig noise.

Then a quasi-real profile, modeling the k8s deployment where this bug actually bites: 150 service-link-style env vars on the server process (`SVC_N_SERVICE_HOST` / `_PORT` / `_TCP_PROTO`) and 20 realistic headers on every request (authorization bearer, cookie with 4 cookies, x-forwarded-*, traceparent, sec-fetch-*, ...). Same `b1k`, c=64, 10 interleaved rounds:

| variant         | mean rps | sd   | vs master  | per-round values |
| --------------- | -------- | ---- | ---------- | ---------------- |
| master          | 55.3k    | 7.2% | -          | 56.4k, 57.1k, 51.2k, 56.9k, 55.6k, 49.0k, 58.1k, 58.8k, 49.7k, 60.3k |
| `setdefault`    | 44.1k    | 1.9% | **-20.2%** | 44.7k, 42.5k, 43.9k, 43.5k, 43.8k, 44.1k, 44.1k, 45.7k, 44.5k, 44.7k |
| `union`         | 54.3k    | 6.8% | -1.8%      | 59.1k, 52.7k, 59.5k, 53.7k, 50.3k, 52.3k, 58.4k, 56.6k, 50.1k, 50.7k |
| union + log fix | 56.0k    | 6.6% | +1.2%      | 58.1k, 52.2k, 56.0k, 50.9k, 58.7k, 58.8k, 49.7k, 59.8k, 58.9k, 56.9k |

The merge step in isolation for that same shape (32-key scope, 150-var env, same CPU): `update` 3.7 us, `setdefault` loop 8.5 us, union 1.8 us. The +4.7 us delta over `update` on an ~18 us request budget predicts the measured -20% almost exactly, and the 1.9% sd (vs 7% for the others) says it's systematic, not noise.

So the setdefault loop is out - I proposed it for being allocation-free, but it's a -20% hit on exactly the deployment class this fix is about. The union is at parity with `update` on both profiles and both benchmarks: one right-sized dict per request turns out cheaper than re-inserting the whole environment into the request dict on every request.

I'll rework the PR to the union + log fix diff above: precedence fixed, access log byte-identical to master, parity on the default profile, parity on the k8s-shaped one. The startup warning/error offer from the previous comment still stands if you'd rather not change the precedence at all.

gi0baro (approved):

Issue #613, opened 2025-06-20
Setting to send ping in websocket connections


saschnet:
I would like to migrate from uvicorn to granian but cannot find a feature for websocket connections.

In uvicorn there is the setting `--ws-ping-interval` with determines the interval the server creates a ping request and sends it to the connected client.

I cannot find an equivalent setting and see that connections are timing out when clients do not implement a ping pong mechanism on their own.

Ideally, the setting would work as in uvicorn. When set to None or 0, there is no ping initiated, otherwise the ping will be triggered in the interval as configured.

When implementing the feature, I guess it also makes sense to be able to config something like `--ws-ping-timeout` in uvicorn.

Keep up great work!

noahnut:
Hi, Can I take this issue ? Thank you

gi0baro:
> Hi, Can I take this issue ? Thank you

Sure thing

saschnet:
Is there any progress on this or is it stale @noahnut ?

gi0baro:
👍 reassigned

saschnet:
Hi @JaeHyuckSa, can you share any news on the progress on this feature?

XChikuX:
@gi0baro Can I pick this up please?

I'm in dire need of this feature.

gi0baro:
@XChikuX assigned to you

PR #882, opened 2026-07-14
Websocket ping

KonradSwierczynski:
Implement https://github.com/emmett-framework/granian/issues/613

### DIFF
--- a/granian/_granian.pyi
+++ b/granian/_granian.pyi
@@ -80,6 +80,9 @@ class ASGIWorker:
         ssl_ca: str | None,
         ssl_crl: list[str],
         ssl_client_verify: bool,
+        metrics: tuple[int | None, Any | None] = ...,
+        websocket_ping_interval_ms: int | None = ...,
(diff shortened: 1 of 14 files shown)


gi0baro:
@KonradSwierczynski while I appreciate the initiative, I'd say rewriting the entirety of the websocket I/O protocol logic, *just to add ping support* is a bad idea, and definitely not the way to go to add support for #613.

Ideally the change should be **way more self-contained**.

Also fyi, in general I don't welcome claude generated code in Granian.

KonradSwierczynski:
> @KonradSwierczynski while I appreciate the initiative, I'd say rewriting the entirety of the websocket I/O protocol logic, _just to add ping support_ is a bad idea, and definitely not the way to go to add support for #613.
> 
> Ideally the change should be **way more self-contained**.
> 
> Also fyi, in general I don't welcome claude generated code in Granian.

Hey, thanks for checking. I wanted to see what the change would look like, but it was bigger than expected, so I didn't want to publish it

PR #887, opened 2026-07-28
Fix memory spikes in streamed WSGI/ASGI/RSGI responses

### LINKED ISSUE #885: Memory spikes when serving files through WSGI
# Problem: Memory consumption when serving large files

Recently I've started experimenting with running [InvenioRDM](https://inveniordm.docs.cern.ch/) (which is a Flask application and a research data repository basically for sharing files) on free-threading Python 3.14t, replacing `uWSGI` with Granian.
This went mostly fine, with one exception: The serving of large (non-static) files through Python/Granian causes memory usage in Granian to spike significantly.

From what it looks like, this has caused the OOM killer to kick in and e.g. kill Granian, causing connections to be dropped.

## Example scenario

I'm running the entire dev setup for the application locally, and in an idle state my PC sits at 6GB RAM usage.
Then I start downloading one of the uploaded files, which is 10GB in size.
As soon as the download starts, the RAM consumption starts spiking up to 16.9GB and then slowly dropping down again.

When the application is run with `Gunicorn` or the Flask development server, this spike does not happen; the memory usage increase is minimal, barely different from the idle state.

## Example Flask application

The following is a small example application to reproduce the behavior.
It's what InvenioRDM does with files that are served through Python; probably not all HTTP headers are required to reproduce, but some of them seem to be.

```python
# file: app.py

from flask import Flask
from werkzeug.datastructures import Headers
import os
import io

app = Flask(__name__)

@app.route("/file")
def serve_file():
    # Point to a large file (e.g. 10GiB)
    filepath = "/path/to/large/file"
    name = os.path.basename(filepath)
    size = os.path.getsize(filepath)

    # Set some headers that are relevant for forcing the file to be downloaded
    # rather than rendered
    headers = Headers()
    headers["Content-Length"] = size
    headers['Content-Security-Policy'] = "default-src 'none';"
    headers['X-Content-Type-Options'] = 'nosniff'
    headers['X-Download-Options'] = 'noopen'
    headers['X-Permitted-Cross-Domain-Policies'] = 'none'
    headers['X-Frame-Options'] = 'deny'
    headers['X-XSS-Protection'] = '1; mode=block'
    headers['Content-Disposition'] = f'attachment; filename={name}'
    headers['Content-Type'] = 'application/octet-stream'
    headers['Last-Modified'] = 'Mon, 27 Jul 2026 08:34:34 GMT'
    headers['Cache-Control'] = 'public'

    f = io.open(filepath, "rb", buffering=-1)
    resp = app.response_class(f, headers=headers, direct_passthrough=True)
    return resp
```

## Run commands

`PYTHON_GIL=0 granian --interface wsgi app --host 0.0.0.0 --port 5000 --ssl-certificate test.crt --ssl-keyfile test.key`

`gunicorn app:app --bind 0.0.0.0:5000 --certfile test.crt --keyfile test.key`

`flask run -h 0.0.0.0 -p 5000 --cert test.crt --key test.key`

## Further information

I've tested the behavior with free-threading Python `3.14t`, as well as "normal" `3.14` and it is consistent across both.
Granian spikes memory consumption, while the Flask development server and Gunicorn don't.

BitWeaverDev:
## Summary

Fixes #885.

Streamed response bodies (WSGI iterables, ASGI `http.response.body` streaming, RSGI `response_stream`) are handed to the hyper transport over an `mpsc` channel. It was unbounded, and the producer side did a fire-and-forget send, so a producer that outpaces the client (e.g. reading a file from page cache at GB/s while a client drains the socket at MB/s) buffered the entire gap in memory.

- `src/wsgi/io.rs`: `response_iter` already runs on a dedicated blocking OS thread (`spawn_blocking`), so it now does a real `blocking_send` with the GIL released (`py.detach`) instead of a non-blocking `send`.
- `src/asgi/io.rs` / `src/rsgi/io.rs`: the send methods are already `await`ed from Python (ASGI `send()`, RSGI `send_bytes`/`send_str`), but the Rust side resolved the returned future immediately without ever honoring channel capacity. They now genuinely `await` the bounded send, so an already-`await`ed call finally does what it always looked like it did.
- All three now share a bounded channel (32 slots) built by a new `crate::http::body_stream_channel()` helper, replacing four near-identical unbounded-channel-construction blocks that were duplicated across the three protocol modules.

## Test plan

- `cargo fmt --all -- --check`, `cargo clippy --tests` with the project's `lint-rust` flags: clean.
- Full `pytest` suite: 176 passed, 0 failed, 4 skipped (unchanged from master).
- Manual repro: served an 800MB file over WSGI to a client throttled to 20MB/s via `curl --limit-rate`, sampling worker RSS:
  - Before this change: **~866MB** peak RSS (essentially the whole file buffered).
  - After this change: **~118MB** peak RSS.

### DIFF
--- a/src/asgi/io.rs
+++ b/src/asgi/io.rs
@@ -23,7 +23,7 @@ use super::{
 };
 use crate::{
     conversion::FutureResultToPy,
-    http::{HTTPResponse, HTTPResponseBody, HV_SERVER, response_404},
+    http::{HTTPResponse, HTTPResponseBody, HV_SERVER, body_stream_channel, response_404},
     runtime::{
         Runtime, RuntimeRef, done_future_into_py, empty_future_into_py, err_future_into_py, future_into_py_futlike,
     },
@@ -43,7 +43,7 @@ pub(crate) struct ASGIHTTPProtocol {
     response_started: atomic::AtomicBool,
     response_chunked: atomic::AtomicBool,
     response_intent: Mutex<Option<(u16, HeaderMap)>>,
-    body_tx: Mutex<Option<mpsc::UnboundedSender<body::Bytes>>>,
+    body_tx: Mutex<Option<mpsc::Sender<body::Bytes>>>,
     flow_rx_exhausted: Arc<atomic::AtomicBool>,
     flow_rx_closed: Arc<atomic::AtomicBool>,
     flow_tx_waiter: Arc<Notify>,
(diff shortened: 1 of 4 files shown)


gi0baro (commented):
@BitWeaverDev thank for your efforts.

Unfortunately I cannot accept this in its current state as:
- there's no rationale for the *empirical 32 slots* number
- this causes a performance regression (measured on throughput vs current master) on all protocols (RSGI -7%, ASGI -20%, WSGI -40%)

I'll probably open a side PR based on this that tackles both point, rather thank asking you to continue the work on this.

gi0baro:
Superseded by #898

Issue #875, opened 2026-07-03
Worker hangs forever on SIGTERM when the event loop is busy


Malkiz223:
## Expected behavior

A worker that received SIGTERM stops gracefully: it finishes serving its connections and closes them, the master logs `Stopped worker-N`, the process exits within milliseconds.

## Actual behavior

When the worker's event loop is busy at the moment of shutdown (real production traffic), the worker can hang forever - at the very end of the shutdown: it has already stopped accepting requests, served and closed everything it had, and hangs on the last step before the process exit. With the default `workers_kill_timeout` (none) the master waits for it forever - the whole server shutdown freezes dead.

What this leads to, in the two common cases:

- Server stop (deploy, `docker stop`). The shutdown never finishes on its own, so after 10 seconds (the default `docker stop` timeout) docker kills everything with SIGKILL. During those seconds the port is open but nobody answers: new connections just wait, and the SIGKILL then breaks them all at once. The reverse proxy in front (nginx in our case) sees the broken connections, marks the node as failed and returns 502 to clients.
- Worker recycle (`workers_lifetime`). From the outside everything looks fine: the new worker starts first and serves the traffic. But the old worker never dies - it stays as a zombie with its memory and connections, the master gets stuck waiting for it, and all future recycles stop happening.

The hang rate depends on host load: on a loaded host it fires on almost every SIGTERM, on an idle machine only occasionally. For us this is not a rare event but a daily one: with `workers_lifetime` recycles and regular deploys we hit it every day. It can look like a rare flaky failure, but it is a genuine permanent deadlock, the race analysis is below.

## Reproduction

In real life the race window is only a few microseconds wide, but every worker stop is a new try, and on a busy host it hits almost every time. The script below makes that same window 0.2 seconds wide, so the hang fires on every run on any machine: `time.sleep` inside `_write_to_self` releases the GIL the same way the socket `send` there does, just for longer.

```python
# app.py
import asyncio
import time


async def keep_loop_busy():
    while True:
        await asyncio.sleep(0)


def make_race_window_bigger(loop):
    inner_write = loop._write_to_self

    def slow_write_to_self():
        time.sleep(0.2)
        inner_write()

    loop._write_to_self = slow_write_to_self


async def app(scope, receive, send):
    if scope['type'] == 'lifespan':
        while True:
            message = await receive()
            if message['type'] == 'lifespan.startup':
                loop = asyncio.get_running_loop()
                loop.create_task(keep_loop_busy())
[...]

PR #901, opened 2026-08-03
Fix lost condvar wakeup on mt worker shutdown

### LINKED ISSUE #900: Worker hangs on shutdown after #899: lost condvar wakeup in the mt serve path
Follow-up to #875. The GIL deadlock is gone, thanks. The regression test from #876 passes on master now.

But I ran the same reproduction from #875 on master, and the worker still hangs on shutdown. The old worker is still alive 20 seconds after its SIGTERM, and then granian SIGKILLs it on `workers-kill-timeout`. The cause is a different one this time, and it came in with #899.

It is only on master. No released version is affected, so there is time to fix it before 2.8.

## The bug

In the `mt` arm of `src/serve.rs` the shutdown thread waits on a condvar without checking the flag:

```rust
let (lock, cvar) = &*cvar;
let done = lock.lock().unwrap();
let _done = cvar.wait(done);
```

The runtime task does `*done = true; cvar.notify_one();`. If it gets there first, the notification is lost, because `std::sync::Condvar` does not store one. The thread then sleeps forever.

After that `release()` is never called, `evp` is never set, and `loop.run_until_complete(_main())` in `mp.py` never returns.

The window is between `stx.send(true)` and taking the mutex. It is short, so this needs CPU pressure to show up. On an idle machine the shutdown thread almost always gets there first.

## The fix

```rust
let mut done = lock.lock().unwrap();
while !*done {
    done = cvar.wait(done).unwrap();
}
```

Sent as a PR.

## A hung worker

py-spy on the main thread, 20 seconds after the SIGTERM:

```
Thread 55 (idle): "MainThread"
    epoll_wait (libc.so.6)
    select (selectors.py:468)
    run_forever (asyncio/base_events.py:645)
    run_until_complete (asyncio/base_events.py:678)
    _spawn_asgi_lifespan_worker (granian/server/mp.py:249)   <- the wait added in #899
```

It sits there and nobody is going to set that event.

At that moment `/proc` still lists all eight `tokio-rt-worker` threads of the worker, so the runtime was never dropped and the wait never returned.

## How I checked

Plain ASGI app, granian recycles the workers itself via `--workers-lifetime`, container with `--cpus=2`. A hang means the old worker is still alive 20 seconds after its own SIGTERM. I check the process, not the log line.

| build | result |
| --- | --- |
| master as is | hung in every run, first hang between 105s and 795s |
| master plus the flag check | nothing in 384 restarts, eight runs of 15 minutes |

To be sure it is really this spot and not something next to it, I also built master with a three second sleep between `stx.send(true)` and taking the mutex, changing nothing else. With it, the first restart hung. With the same sleep and the flag check, none of 40 restarts hung.

Same setup as in #875, nothing new was needed. Happy to share the reproduction scripts.

Malkiz223:
Fixes the worker hanging on shutdown in the `mt` arm of `src/serve.rs`.

The shutdown thread waited on the condvar without checking the flag. The runtime task sets the flag and calls `notify_one()`. If it gets there first, the notification is lost, because `std::sync::Condvar` does not store one. The thread then sleeps forever, `release()` is never called, the Python loop never returns, and granian SIGKILLs the worker on `workers-kill-timeout`.

It now waits in a loop over the flag. `wait_while` would do the same in one line if you prefer that.

Without the fix the reproduction from #875 hung in every run, first hang between 105s and 795s. With it, nothing in 384 restarts over eight runs of 15 minutes.

I also built the branch with a three second sleep in that window to widen it. Without the fix the first restart hung, with the fix none of 40 did.

Fixes #900, the full write-up is there.

### DIFF
--- a/src/serve.rs
+++ b/src/serve.rs
@@ -107,8 +107,10 @@ macro_rules! serve_fn {
                 _ = stx.send(true);
 
                 let (lock, cvar) = &*cvar;
-                let done = lock.lock().unwrap();
-                let _done = cvar.wait(done);
+                let mut done = lock.lock().unwrap();
+                while !*done {
+                    done = cvar.wait(done).unwrap();
+                }
 
                 Python::attach(|py| {
                     _ = pysig.get().release(py);


gi0baro (approved):

gi0baro:
@Malkiz223 thank you for your contribution

PR #902, opened 2026-08-07
Import py-setproctitle right before use to avoid MacOS fork crash

iamkhav:
## Description
There is an issue with py-setproctitle on MacOS where forking a new process and renaming causes a SEGFAULT.

A workaround is described in https://github.com/dvarrazzo/py-setproctitle/issues/127#issuecomment-3770132728: Import py-setproctitle right before using it which circumvents the issue. 

I've tested this on my company laptop which runs MacOS 26 and it works.

Hope Granian is okay with supporting this workaround. I'm open to suggestions on how to make this prettier.

### DIFF
--- a/benchmarks/vs.md
+++ b/benchmarks/vs.md
@@ -136,4 +136,3 @@ Granian is run with `--ws` and `--runtime-threads 2`.
 | 32 | Gunicorn Asgi | 3802710 | 366978 | 378446 |
 | 32 | Uvicorn H11 | 3163470 | 183706 | 189447 |
 | 32 | Hypercorn | N/A | N/A | N/A |
-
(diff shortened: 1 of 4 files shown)


gi0baro:
@iamkhav given that `fork` is not the default multiprocessing spawn method on MacOS, and the issue is clearly upstream, I'm not sold on the idea this should be handled in granian by making its code worse..

iamkhav:
> @iamkhav given that `fork` is not the default multiprocessing spawn method on MacOS, and the issue is clearly upstream, I'm not sold on the idea this should be handled in granian by making its code worse..

Yeah, that's a valid statement. I'll close this PR.

I'm going to open an issue for the following piece of logic:
https://github.com/emmett-framework/granian/blob/09e1a8a94bb19fee7e8447884f3a541caa7a1bda/granian/server/common.py#L680-L682
The mere presence of `setproctitle` in the venv causes Granian to set a process title, causing an issue when using setups that are unstable/touchy with that lib. I'm hoping we could at least connect that to `process_name` explicitly being given.

Also, someone on the [issue I had opened](https://github.com/dvarrazzo/py-setproctitle/issues/165#issuecomment-5180405698) for the upstream library suggested [their own alternative](https://github.com/gershnik/processtitle). This circumvents the problem. Not sure if you'd like that but the claim is that it's layout is improved over py-setproctitle. 

Thanks for your time.

PR #904, opened 2026-08-10
Fix lost shutdown wakeup for connection tasks

### LINKED ISSUE #903: Worker still hangs on shutdown: connection tasks can miss the stop signal
Follow-up to #875 and #900. The fix from #901 works, but a worker can still hang on shutdown, now much more rarely. I ran the same reproduction from #875 against 2.8.0, 2.8.1 and master, and it hangs on all of them, roughly once per a few hundred worker recycles under CPU pressure. Both mt and st are affected, the mechanism is shared.

## The bug

Same class as #900: a notification that nobody stores is lost when the receiver is not waiting yet. Only this time between tasks instead of threads.

On shutdown [the accept loop](https://github.com/emmett-framework/granian/blob/9a20b1742b97ab8e85f58220f98f1d14e35387c3/src/workers.rs#L1077-L1080) signals the connection tasks:

```rust
_ = $sig.changed() => {
    accept_loop = false;
    connsig.notify_waiters();
}
```

[A connection task](https://github.com/emmett-framework/granian/blob/9a20b1742b97ab8e85f58220f98f1d14e35387c3/src/workers.rs#L690-L698) waits for it in a `select!`:

```rust
tokio::select! {
    biased;
    _ = conn.as_mut() => { done = true; },
    _ = $sig.notified() => {
        conn.as_mut().graceful_shutdown();
    }
}
```

`notify_waiters()` only wakes tasks that already reached `notified()`, and it stores nothing. A task that `tasks.spawn(...)` has created but the scheduler has not run yet loses the signal for good. It then waits on `notified()`, which will never come. A keep-alive connection does not close on its own while the client is alive, so [`tasks.wait().await`](https://github.com/emmett-framework/granian/blob/9a20b1742b97ab8e85f58220f98f1d14e35387c3/src/serve.rs#L91-L92) never returns, `release()` is never called, and the worker sits there until `workers-kill-timeout`.

The window between spawning a task and its first poll is normally short, which is why the bug is rare and needs CPU pressure. The `biased` in the accept loop widens it: when a new connection and the stop signal are ready at the same time, the task is spawned first, and the notification goes out only on the next loop iteration, when that task has almost certainly not run yet.

## The fix

The accept loop already has a level-triggered stop channel, the `watch`. I hand every connection task a clone of it instead of the `Notify` and wait via `wait_for`, which returns right away when the value is already set:

```rust
_ = $sig.wait_for(|v| *v) => {
    conn.as_mut().graceful_shutdown();
}
```

## How I checked

Same setup as in #875, only the workers recycle faster. A hang means the old worker is still alive 20 seconds after its own SIGTERM.

| build | result |
|---|---|
| 2.8.0 | 2 hangs in 1044 recycles |
| 2.8.1 | 2 hangs in 692 recycles |
| master + a 3s sleep at the top of the connection task | 13 hangs in 56 recycles |
| master + the same sleep + the fix | 0 in 136 |
| master + the fix, no sleep | 0 in 1152, an hour each in mt and st |

The sleep sits before the `select!`, it only stretches that exact window and changes nothing else. Without the fix every fourth recycle hangs, with it none do.

py-spy of a hung worker looks the same as in #900, the main thread sits in `run_until_complete` in `mp.py` waiting for an event nobody is going to set.

Happy to share the reproduction scripts.

Malkiz223:
Fixes the worker hanging on shutdown, shared between mt and st.

Connection tasks learned about shutdown from a `Notify` signalled with `notify_waiters()`. That call only wakes tasks already parked in `notified()`, and it stores nothing. A task that was spawned but has not run yet lost the signal for good. With a keep-alive client its connection never closes on its own, `tasks.wait()` never returns, and the worker sits there until `workers-kill-timeout`.

Every connection task now gets a clone of the already existing shutdown `watch` channel and waits via `wait_for`, which also returns when the value is already set.

Fixes #903, the full write-up and the numbers are there.

### DIFF
--- a/src/workers.rs
+++ b/src/workers.rs
@@ -677,7 +677,7 @@ trait WorkerHandle<I, S> {
         svc: S,
         stream: I,
         permit: tokio::sync::OwnedSemaphorePermit,
-        sig: Arc<tokio::sync::Notify>,
+        sig: tokio::sync::watch::Receiver<bool>,
     ) -> impl Future<Output = ()> + Send + 'static;
 }
 
@@ -692,7 +692,7 @@ macro_rules! conn_handle_h1_impl {
             _ = conn.as_mut() => {
                 done = true;
             },
-            _ = $sig.notified() => {
+            _ = $sig.wait_for(|v| *v) => {
                 conn.as_mut().graceful_shutdown();
             }
         }
@@ -736,7 +736,7 @@ macro_rules! conn_handle_ha_impl {
             _ = conn.as_mut() => {
                 done = true;
             },
-            _ = $sig.notified() => {
+            _ = $sig.wait_for(|v| *v) => {
                 conn.as_mut().graceful_shutdown();
             }
         }
@@ -771,7 +771,7 @@ macro_rules! conn_handle_h2_impl {
             _ = conn.as_mut() => {
                 done = true;
             },
-            () = $sig.notified() => {
+            _ = $sig.wait_for(|v| *v) => {
                 conn.as_mut().graceful_shutdown();
             }
         }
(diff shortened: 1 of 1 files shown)


gi0baro:
Superseded by #910

PR #924, opened 2026-09-23
Add WebSocket ping/pong keepalive configuration

### LINKED ISSUE #613: Setting to send ping in websocket connections

XChikuX:
Closes #613
 
## Summary
 
Adds server-initiated WebSocket keepalive with uvicorn-equivalent semantics, behind two new CLI/env options:
 
| Option | Env var | Default | Meaning |
|---|---|---|---|
| `--ws-ping-interval` | `GRANIAN_WS_PING_INTERVAL` | `None` (disabled) | Seconds between Ping frames sent to each connected client. `0` disables |
| `--ws-ping-timeout` | `GRANIAN_WS_PING_TIMEOUT` | `60.0` | Seconds to wait for the matching Pong before closing the connection. `0` disables the timeout (pings without enforcement) |
 
```sh
granian --interface asgi app:app --ws-ping-interval 7.5 --ws-ping-timeout 30
```
 
The options are exposed in the CLI through `Server` (multiprocess, threaded, and embedded) into the Rust worker config, and apply to both **ASGI** and **RSGI** applications in every runtime mode.
 
## How it works
 
- Keepalive is opt-in: with the default `None` interval nothing changes — no task is spawned, no frames are sent.
- When enabled, each accepted WebSocket connection spawns a per-connection keepalive task that sends a Ping every `interval` seconds, carrying a unique 4-byte payload.
- Pongs are matched against the ping currently in flight; stale or unsolicited Pongs are ignored. Accounting is a lock-free atomic + notify pair, so the hot receive path is untouched.
- If no matching Pong arrives within `timeout`, the server closes the connection with close code `1011` / reason `keepalive ping timeout`, tears down the transport, and unblocks any pending `receive()` as a disconnect — reusing the existing `disconnect_guard` machinery.
- Client-initiated Pings continue to be answered automatically by the protocol layer, unchanged.

## Compatibility
 
- No new dependencies; no changes to the WebSocket upgrade path, protocol handling, or shutdown semantics.
- Non-WebSocket paths (WSGI, metrics, IPC) carry a disabled config and are unaffected.

## Tests
 
New coverage in `tests/test_ws.py` (parametrized over asgi/rsgi × mt/st):
 
- `test_ping_frame_sent` — server emits a Ping frame (opcode `0x9`, 4-byte payload)
- `test_ping_timeout_closes_connection` — a peer that never pongs is closed with `1011`
- `test_ping_disabled` — no Pings are sent when the interval is unset or `0`
- `test_ping_interval_without_timeout` — pings are sent but no timeout is enforced when `--ws-ping-timeout 0`
- `test_keepalive_survives_pong` — a client that answers Pings stays connected past the timeout window
Plus `tests/test_cli.py::test_ws_ping_options_exposed`, pinning option defaults and env-var exposure.

> I have also run this on my live server and faced no glaring issues.
 
## Docs
 
- README: options list and a new paragraph in the Websockets section covering semantics and the uvicorn mapping
- CLI help text states the `0`-disables semantics for both flags
- `granian/_granian.pyi` stub updated

## Future work:
 
A small follow-up is planned to observe Pongs independently of the application's read loop (a dedicated reader task feeding a bounded queue), so push-only apps that never call `receive()` get the same keepalive guarantees with bounded buffering under client floods.


gi0baro (commented):
Ain't no way I can accept this.

Few things:
- if I wanted an LLM to implement this, I would've prompted it myself
- there's no reason for this to touch `runtime.rs`, `metrics.rs` and `ipc.rs`
- 550 lines for this is just unacceptable and a liability I don't want to pay forever

XChikuX:
I checked what it added at every step. They were WS config pass downs. Seemed fairly minimal to me.

I'm afraid LLMs do better rust code than myself at this stage. Happy to edit manually based on your review.

The main change is at `ws.rs`. Every other file has been touched hyper minimally only.

gi0baro:
> Seemed fairly minimal to me.

The point is not about "how much code changed" there, it's about things that shouldn't be touched to deal with websockets.

> LLMs do better rust code than myself at this stage

The problem is not the code itself, it's the lack of architectural knowledge to make decent decisions.
Again, there's no room in Granian for introducing changes that nobody understand.
What I'm saying is that it's not about "cleaning this PR implementation", it's "this PR is fundamentally garbage" and it makes no sense to spend more time (should I say tokens?) dealing with it.

tests/test_rsgi.py
XChikuX: This is a real bug in your tests. Please fix, even if this PR doesn't land

XChikuX:
Most of the LOC's are 

```
        'ws_ping_interval': ws_ping_interval,
        'ws_ping_timeout': ws_ping_timeout,
```

And `keepalive` additions. It's really not that hard to understand. I only responded with what I did cause you mentioned `550 lines`

> lack of architectural knowledge to make decent decisions.

I am not your typical `vibe coder`. I have extensive experience in system design (full protocol design) and have been completely skeptical of AI use since the very beginning. I have had real gun to the foot moments and have seen the full evolution of where I can trust and not trust the outputs.

There is no way someone enters into this codebase without your guidance and makes changes that are up to mark. No contributor should be expected to know the full codebase to contribute.

I personally checked every inch of this code with your surrounding logic and determined that this feature was implemented satisfactorily. So unless you have issues with `ws.rs` I don't see any reason to really **at least** not give this a chance.

If you are concerned about upkeep I can personally devote my time to this specific `ping_interval` logic. I am more than qualified. I am unfortunately working on several things and LLM use really helps me get in changes in repo's that I highly respect and use and don't have infinite time to spend on.

>tldr;

This was not an attempt to push in garbage is what I'm trying to say.

gi0baro:
> There is no way someone enters into this codebase without your guidance and makes changes that are up to mark. No contributor should be expected to know the full codebase to contribute.

I never expected that. But asking me in advance is free. That's what issues and my public email are here for.

> I personally checked every inch of this code with your surrounding logic and determined that this feature was implemented satisfactorily. So unless you have issues with ws.rs I don't see any reason to really at least not give this a chance.

Not for me. And I stated what's obviously wrong in my first message.
If you want to spend more tokens in reshaping the code to address those comments, then go ahead, I can then look at the actual implementation in `ws.rs` and `io.rs`. But you cannot expect me to spend my time reviewing generated code from someone else with the same approach/effort I would do over a genuine implementation.

